PR #4, opened 2023-07-11
Network nodes

annehaley:
This PR does the following:
1. adds a new django model called `NetworkNode`
2. adds a boolean flag called `network` to the `Dataset` model
3. adds a function in the vector data conversion task that can create `NetworkNode` objects for a dataset where network=true


annehaley:
The process for creating a network representation of a dataset is as follows:
1. Create a geodataframe representation of the dataset (this is something we already did in the conversion task to write a geojson representation)
2. Separate the geodataframe into a node set and an edge set
3. Group edges by common values in a certain column known as the `connection_column` ("ROUTE" for the subway dataset and "LINE_BRNCH" for the commuter rail dataset; these are set in the dataset's options).
4. For each edge group, create a single route geometry from the combination of all the edge segments in that group
5. For each edge group, consider all the nodes that share a portion of the value for the `connection_column`. These nodes are expected to be on that route. Along all the points of the single route geometry, find the closest node to each of those points. Condense this to a list of unique values. The resulting node list is in the order that follows the single route geometry from starting point to ending point.
6. Use the ordered list of nodes along the route to record adjacencies
7. Create a `NetworkNode` object for every node in the node list. Record adjacencies.

annehaley:
There are some limitations to this algorithm due to faults/idiosyncrasies in the input data. These come in two categories. 

The first is poor labeling on the nodes (stations). For example, South Station should be connected to Downtown Crossing, Broadway, Courthouse, and Chinatown according to the official MBTA map below. It should be on both the Silver line and the Red line. But in the input dataset, the row for South Station has "SL1 SL2 SL3" as the value for "ROUTE". With this algorithm, the resulting connections are just on the silver line. This may be unavoidable unless we overwrite the source data.

image

The second problem with the data is strange line geometries. Below is an example segment of the Silver line, which runs from Chinatown to Nubian. The actual geometry in the input data is not in the correct order. The line starts in the middle, where the marker is. Then it goes down to Nubian and back up again. The algorithm therefore finds the stations along this route in the wrong order.  

image

I think this second problem can be solved with an additional step to process the geometry, but this would require non-negligible effort. We would need to deconstruct the line into its points and then find the shortest path from those points. But this is a big problem with no existing builtin function we can leverage. See below link for possible solution.

https://gis.stackexchange.com/questions/420940/shortest-path-touching-all-points

annehaley:
I will make a separate PR for displaying the NetworkNodes on the map. This PR is scoped only for the backend.

johnkit (approved):
Excellent work. Go ahead and merge.

PR #5, opened 2023-07-19
Add more colormap options and use terrain as default

annehaley:


johnkit (approved):
Looks good let's merge.

PR #6, opened 2023-07-21
Network vis

annehaley:
This PR adds a toggle switch to the Options panel for any dataset where `network=True`. 
When enabled, the normal layer is hidden and a new layer with the network representation is created (or shown if it already exists). For now, this visualization is static and does not have special interaction options (such as marking nodes as inactive). It does have the tooltip interaction to view the details of a node.

image

### DIFF
--- a/uvdat/core/serializers.py
+++ b/uvdat/core/serializers.py
@@ -18,8 +18,6 @@ class NetworkNodeSerializer(serializers.ModelSerializer):
 
 
 class DatasetSerializer(serializers.ModelSerializer):
-    network_nodes = NetworkNodeSerializer(many=True)
-
     class Meta:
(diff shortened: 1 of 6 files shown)


johnkit:
That looks great. I hadn't thought about this before, but is it straightforward to apply the same color scheme to the lines and stations that you use for the GIS data?

annehaley:
> That looks great. I hadn't thought about this before, but is it straightforward to apply the same color scheme to the lines and stations that you use for the GIS data?

Yes, we can actually use the same `createStyle` function. I had just gone with b/w to visually distinguish between the modes. But either way works.

johnkit (approved):
Looks real good. Let's leave the coloring as-is for now.

PR #7, opened 2023-07-24
Tide level dataset

annehaley:
This PR does the following:

1. Add a new field `metadata` to the Dataset model. This field is a JSON and can hold any miscellaneous relevant information for the dataset. We can find a way to display this in the UI later.
2. Add the tide level dataset to `datasets.json`. This dataset has a chunk of information saved to the `metadata` field. It has a new value for `raw_data_type`: `csv` and a new value for `category`: `sensor`. (Note: since the `raw_data_type` is `csv`, the conversion task is a no-op)
3. Use the new `metadata` field for some things that had been in the `style` field. I had been using the `style` field for specifying certain column information about network datasets, and this had felt like I was overloading the `style` field. The `metadata` field, now that we have it, is more appropriate for this.


johnkit (approved):
Let's go!

PR #11, opened 2023-08-09
UI updates

annehaley:
This PR makes the following UI changes:

- Organize available layers by category, using expansion panels to show/hide groups
- Move the options panel to the righthand side of the map to create more visual separation between the two panels
- Use a gear icon to open the options panel, shown in both the active layers list and the available layers list (if a layer is active)
- Add a toggle to the options panel for any raster dataset "Show raster tooltip". When enabled, mousing over the raster will give an approximate value. Values are approximate because the data is received at a 1/10 resolution to reduce load time, and because mouse position to array index conversion is done by proportion of coordinates.
image

### DIFF
--- a/uvdat/core/views.py
+++ b/uvdat/core/views.py
@@ -1,9 +1,12 @@
 import json
+import ijson
 import tempfile
+import large_image
+from pathlib import Path
 
 from django.http import HttpResponse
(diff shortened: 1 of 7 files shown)


johnkit (approved):
great work.

PR #12, opened 2023-08-10
Regions

annehaley:
For @AlmightyYakob to take over. The below items are my current understanding of the task, but this may be further refined.

- [x] Create Region model
- [x] Populate Region objects from the Boston census blocks dataset
- [x] Add a UI widget to select a Region in the current city and crop the visualization to the boundary
- [ ] Allow more complex spatial queries in the widget to combine Regions (union, intersection)
- [ ] Allow user to save Region combinations as new named Region objects


AlmightyYakob:
@annehaley I'd like to address the last 2 items of that checklist in a follow-up PR. The current change is pretty well scoped, and those last two items could involve a decently large change themselves.

annehaley:
> @annehaley I'd like to address the last 2 items of that checklist in a follow-up PR. The current change is pretty well scoped, and those last two items could involve a decently large change themselves.

Makes sense. I can review this PR as-is later today.

annehaley:
I tried out the zoom to region feature a lot, and it works fine for most regions. The only one I found that did not work as expected was the Jamaica Plain region. This is what it zoomed to:
image

AlmightyYakob:
It seems that both of those polygons are a part of the same region. See below where I clicked on the "main" part of it. You'll notice that in both your screenshot and in mine, the map zooms to fit both regions. 

image


Both areas are connected by a thin strip that runs along the boundary for "Longwood". I tried to verify that this is the correct boundary and wasn't a data ingress error by loading the Jamaica Plain feature into a separate online viewer, but it doesn't render it for some reason. The screenshot below shows that thin connecting piece

image

annehaley:
Oh ok, thanks for clarifying! It seems to be working properly then! I had clicked on the tiny portion and thought that was the whole Region. Perhaps this is a good reason to go back to trying to color them uniquely... But that can be saved for later.

uvdat/core/views.py
```diff
@@ -27,6 +27,16 @@ class DatasetViewSet(ModelViewSet, LargeImageFileDetailMixin):
...
+    @action(detail=True, methods=['get'])
+    def regions(self, request, **kwargs):
+        dataset = self.get_object()
+        if dataset.category != 'region':
+            return HttpResponse('Not a region dataset', status=400)
+
+        # Serialize all regions as a feature collection
+        multipolygons = Region.objects.filter(dataset=dataset)
+        return HttpResponse(serialize('geojson', multipolygons, geometry_field='boundary'))
+
```
annehaley, line 39: Since we were talking about the design of the rest endpoint, perhaps it makes more sense to make a new viewset just for Regions, rather than making it a function on either the Dataset or City viewsets. That way we can query Regions however we need to, whether that be by Dataset relation, City relation, or spatial. What do you think?

AlmightyYakob: I think once we're querying and/or creating regions in multiple ways, we should re-evaluate, and move more towards a configuration supporting multiple relations (dataset, city, spacial, etc.). Before then (and before we decide on how we're going to implement combined / user created regions), I'd be essentially guessing as to how to best lay out the endpoints. Since I'd also like to avoid creating a Region viewset as a "placeholder" (to prevent clutter/confusion), I think for now keeping this is fine. What do you think?

annehaley: Yeah I can agree to keeping it there for now, as long we stay open to redesign in future iterations. Perhaps I was thinking too far ahead. It's true that we don't need anything robust at the moment.

annehaley: I'd approve the PR now, but git won't let me because I started it. You can go ahead and merge, though.

AlmightyYakob (approved):

PR #13, opened 2023-08-18
Charts

annehaley:
This PR depends on #12 because both include new migration files, which should be applied sequentially.

I realized that chart data is different enough from map data that a new model should be created to represent them, so this PR does the following:

Server side
- [x] Add a new model `Chart`
- [x] Move the tide level data to a new file `charts.json` for population as a `Chart` object instead of a `Dataset` object
- [x] Add tabs in the top of the UVDAT main drawer in the UI, allowing the user to toggle between viewing modes. Currently there are two tabs, "Map" and "Chart"
- [x] Add functions to create and append to a `Chart` object any time the `compute_gcc` rest endpoint is accessed. The resulting chart data shows a point for every time gcc is computed.

Client side
- [x] Add a ChartJS component to the main content area in the UI, shown when the view mode is switched to "Chart"
- [x] Display a list of available `Chart` objects associated with the current City, shown when the view mode is switched to "Chart"
- [x] Populate the ChartJS component when a `Chart` object is selected
- [x] Add a widget for the user to control X-Axis slicing on the chart when the X range is large (like for the tide level data)
- [x] Add an expansion panel to allow the user to look at the metadata on the selected `Chart` object


annehaley:
image
This is what the GCC chart looks like in the basic use case.  I used a targeted approach when turning nodes off, doing my best to chop the gcc in half. I used a random approach for turning nodes back on. The blue line shows how the gcc changed as I did this; large jumps in the line represent the loss or gain of a critical node.

image
This is what the GCC chart looks like when I oscillate between turning nodes off and turning nodes on. The red line shows my transitioning between adding and subtracting disabled nodes.

annehaley:
image

This is what the Daily Tide Levels chart looks like

johnkit (approved):
looks good outside the box.

web/src/components/MainDrawerContents.vue
```diff
@@ -184,6 +209,29 @@ export default {
         </draggable>
       </v-expansion-panel-text>
     </v-expansion-panel>
+
+    <v-expansion-panel title="Available Charts">
+      <v-expansion-panel-text>
+        <v-icon @click.prevent="fetchCharts" style="float: right"
+          >mdi-refresh</v-icon
+        >
```
AlmightyYakob, line 217: Just an FYI, the positioning of the refresh button makes it seem like it applies to the first item in the list. Maybe some padding or a line underneath this would clear things up?

AlmightyYakob (changes requested):
Looks good overall, just a couple comments

web/src/components/MainDrawerContents.vue
AlmightyYakob: Sorry to further delay, but it seems it's impossible to click the refresh button without opening/closing the chart expansion panel

AlmightyYakob (approved):

PR #15, opened 2023-08-31
Layer Visibility Logic

annehaley:
This PR is intended to fix any unexpected behavior we found with the map layers. This includes the following:

- [x] Prevent disabled layers from turning back on when a network layer is toggled
- [x] Prevent layer reordering when changing which Dataset is open in the right-hand Options Panel
- [x] Clear raster tooltip when a raster Dataset is not open in the Options Panel; the raster tooltip was overriding normal tooltip behavior and preventing the user from selecting a network node

Additionally, a new layer visibility feature was added:

- [x] A map button now appears in the top-right corner to toggle the visibility of the map layer.

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -8,7 +8,9 @@ import {
   loading,
   loadCities,
   activeChart,
+  showMapBaseLayer,
 } from "./store";
+import { updateVisibleLayers } from "./utils";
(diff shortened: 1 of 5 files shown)


johnkit (approved):
Tested all good.

AlmightyYakob (approved):
LGTM

PR #22, opened 2023-09-21
Add allauth account middleware to settings

AlmightyYakob:
After rebuilding my docker image, I'm getting incessant errors about this allauth middleware setting, which prevent Django from running in any capacity.

@annehaley I'm curious to see if this works for you without issue on an existing (not recently rebuilt) local image.

### DIFF
--- a/uvdat/settings.py
+++ b/uvdat/settings.py
@@ -32,6 +32,8 @@ class UvdatMixin(ConfigMixin):
             's3_file_field',
         ]
 
+        configuration.MIDDLEWARE += ['allauth.account.middleware.AccountMiddleware']
+
         configuration.DATABASES = {
             'default': {
                 'ENGINE': 'django.contrib.gis.db.backends.postgis',


annehaley (commented):
We ran into the same problem on ShapeWorks, another Girder 4 project. 

Our first attempt had been to do this exact thing: https://github.com/girder/shapeworks-cloud/pull/339. It looked like it had worked, but it was only behaving properly in our CI; all our deployed instances were broken with this change. 
Instead, we ended up reverting that change and pinning `django-allauth` for the time being (at least until this is fixed upstream): https://github.com/girder/shapeworks-cloud/pull/341.

If you look at the tests, the first PR is all green, so we thought it would work. But then the tests on master failed after merging because our deployed version was broken (and one of our tests tries to communicate with the deployed version).
So David put in the second PR, which looks red because at that point the deployed version was still down. But after merging the second PR, our tests on master all passed and our deployed version was working again.

We don't have a deployment for this project (or tests even), so I don't know if we would ever have this problem. But pinning the version was what worked best on that other project, so it's worth considering. I am open to using either solution on this project since we're only affecting our own dev versions.

AlmightyYakob:
I'm okay with version pinning. If you have a working local image without this issue, what version of allauth is installed?

annehaley:
My container is currently using `0.55.0`, but we pinned to `0.56` on ShapeWorks.

AlmightyYakob:
Thanks for the advice! Closing in favor of #29.

PR #24, opened 2023-09-21
Multi-line Charts

annehaley:


annehaley:
Closing because this was based on the old web client (pre-#34). If this functionality is desired, I will need to re-write it on top of the new design. I can refer to this version to speed up the process, though.

PR #30, opened 2023-10-09
Model redesign

annehaley:
This PR is the first in a series of three PRs in a large redesign effort. This PR is scoped to include changes to database models, their rest endpoints, their conversion functions, and the ingest process that creates objects from our sample data (the `populate` management command).
- [x] Reorganize models 
- [x] Rewrite Rest API for new models 
- [x] Rewrite populate script and tasks to fit creation of new models

Follow-up PRs address the other facets of the redesign effort. Ideally, we should prepare and review them all together so that they can be merged all at once.
- [x] Write testing for CI (#33)
- [x] Adjust web client to fetch and adhere to new models (#34)

image


uvdat/core/models/networks.py
```diff
@@ -0,0 +1,26 @@
...
+        return []
+
+
+class NetworkEdge(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    dataset = models.ForeignKey(Dataset, on_delete=models.CASCADE, related_name='network_edges')
+    metadata = models.JSONField(blank=True, null=True)
+    capacity = models.IntegerField(null=True)
+    line = geo_models.LineStringField()
+    connected_nodes = models.ManyToManyField('NetworkNode')
```
AlmightyYakob, line 26: I think a many-to-many field is overkill for this, and could prevent us from enforcing correctness. It might just be simpler to do the following:

```suggestion
    from_node = models.ForeignKey(NetworkNode, related_name='+', on_delete=models.CASCADE)
    to_node = models.ForeignKey(NetworkNode, related_name='+', on_delete=models.CASCADE)
```

We could also store a field that tells us if the edge is directed or not. I don't know if that makes sense with any use cases we have thus far, but it's certainly a common use case w.r.t. network data:
```
directed = models.BooleanField(default=False)
```

Do you think this will work?

uvdat/core/models/networks.py
```diff
@@ -0,0 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    dataset = models.ForeignKey(Dataset, on_delete=models.CASCADE, related_name='network_nodes')
+    metadata = models.JSONField(blank=True, null=True)
+    capacity = models.IntegerField(null=True)
+    location = geo_models.PointField()
+    connected_edges = models.ManyToManyField('NetworkEdge')
+
+    def get_adjacent_nodes(self):
+        # TODO: get adjacent nodes from connected_edges
+        return []
```
AlmightyYakob, line 17: If the other suggestion about edge modeling is accepted, this could be implemented as such:

```suggestion

    def get_adjacent_nodes(self):
        from_edges = (
            NetworkEdge.objects.filter(to_node=self.id)
            .values_list('from_node_id', flat=True)
            .distinct()
        )
        to_edges = (
            NetworkEdge.objects.filter(from_node=self.id)
            .values_list('to_node_id', flat=True)
            .distinct()
        )

        return NetworkNode.objects.exclude(id=self.id).filter(
            models.Q(id__in=from_edges) | models.Q(id__in=to_edges)
        )
```

uvdat/core/models/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+class Dataset(models.Model):
+    class DatasetType(models.TextChoices):
+        VECTOR = 'VECTOR', 'Vector'
+        RASTER = 'RASTER', 'Raster'
+        CHART = 'CHART', 'Chart'
+
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True, blank=True)
+    city = models.ForeignKey(City, on_delete=models.CASCADE, related_name='datasets')
+    category = models.CharField(max_length=25)
```
AlmightyYakob, line 16: I think we should make this a choice field, like you've already done with `dataset_type`. What do you think?

AlmightyYakob:
> Considering this, developers can use this branch with its new migration history while the docker file refers to the new location, and they can use other branches with their old database location. If this is merged, we will no longer need the old database locations.

The local storage of volume data is meant as a temporary measure for reviewing/testing out this PR, correct (i.e. wouldn't be merged to master)? This configuration makes it otherwise pesky to do a local wipe of the database.

uvdat/core/models/networks.py
annehaley: Yes, I like this concept. I think that we should allow the backwards relation, though (not use "+" for the `related_name`). It would allow us to remove the `connected_edges` field on the `NetworkNode` model. That way it will always be edges referring to nodes.

uvdat/core/models/networks.py
annehaley: Thank you, I will include this.

uvdat/core/models/dataset.py
annehaley: It's tempting, but I don't want us to need a migration every time a user needs to make a new category. I think new values should be allowed for this field, since it doesn't really require any changes in implementation like our other choice fields would. We simply use this field to group datasets in the UI, so a new value in this field would mean a new group, and that would be fine.

annehaley:
> The local storage of volume data is meant as a temporary measure for reviewing/testing out this PR, correct (i.e. wouldn't be merged to master)? This configuration makes it otherwise pesky to do a local wipe of the database.

Sure, this change can be removed before merging. I find it convenient that I can see the volume-mounted folders in the project location, but I understand that's just a personal preference. Besides, it deviates from the Girder 4 setup, so this change can just be scoped to this WIP.

annehaley:
I made another iteration on the design ([aaa1c7f](https://github.com/OpenGeoscience/uvdat/pull/30/commits/aaa1c7fa7fab9518d6daba57da7c1932245a6117)), with the following changes @AlmightyYakob and I have discussed:

- Changing the relationship between `NetworkNode`s and `NetworkEdge`s, as suggested above.
- Removing the inheritance for regions. `OriginalRegion` and `DerivedRegion` just have some repeated fields instead. 
- Additionally, `DerivedRegion` now has a reference to a `VectorDataSource` as its map representation, since it doesn't have a reference to a `Dataset`. We discussed how the region models themselves are internal representations, just like network nodes and edges. For visualization on the map, we should rely on `DataSource` -type objects. Therefore, `DerivedRegion`s need a reference to one of these, which we know should be of the `Vector` type.
- Removing the attempt at a `DataCollection` model. This is a concept we should keep in mind for future iterations, but it doesn't make sense at the moment. This concept is intended for a user to curate their own collections, so this model would likely be tied to a `User` object. At this stage, we disregard `User` objects and have not implemented authentication, so these objects would be largely purposeless. We will need them in the future, though, and they will be easy to add in later, since they are outside of the main web of models.

uvdat/core/models/networks.py
AlmightyYakob: There's no reason a backwards relation is required. You can achieve the same outcome by just using the `NetworkEdge` model:
```python
NetworkEdge.objects.filter(from_node_id=1)
NetworkEdge.objects.filter(to_node_id=4)
```

The reason I opted to not use the backwards relation is because, from past experience, when you have two relations like that, it can get confusing what means what. For example, with the backwards relations like so:

```python
from_node = models.ForeignKey(NetworkNode, related_name='from_edges', on_delete=models.CASCADE)
to_node = models.ForeignKey(NetworkNode, related_name='to_edges', on_delete=models.CASCADE)
```

Let's say the edge table looks something like this:


| from_node  | to_node |
| ------------- | ------------- |
| 1 | 3 |
| 1 | 4 |
| 2  | 4 |

If you wanted to retrieve all of the edges in which node `1` is a _start node_, or wanted to know for which edges node `4` is an _end node_, you would do the following:

```python
NetworkNode.objects.get(id=1).from_edges.all()
NetworkNode.objects.get(id=4).to_edges.all()
```

I don't think it's readily apparent that `from_edges` returns "the edges for which the node is in the starting position". It would be very easy to think `from_edges` actually means "the edges _from_ other nodes into this node", i.e. "the edges in which this node is in the _ending_ position". Reversing the naming scheme has similar problems. (I actually had to re-write this, as the first time I mixed up the two).

I think the usage of `NetworkEdge` to achieve this is much more clear, and prevents confusion (it's also more simple, as now there's only one way to do this kind of query, instead of two or three).

What do you think? If you would still prefer a backwards relation, I think we should try to carefully name them, although I don't see a reason they're actually needed.

uvdat/core/models/networks.py
```diff
@@ -0,0 +1,38 @@
...
+            models.Q(id__in=from_edges) | models.Q(id__in=to_edges)
+        )
+
+
+class NetworkEdge(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    dataset = models.ForeignKey(Dataset, on_delete=models.CASCADE, related_name='network_edges')
+    metadata = models.JSONField(blank=True, null=True)
+    capacity = models.IntegerField(null=True)
+    line = geo_models.LineStringField()
```
AlmightyYakob, line 35: Is this supposed to represent the real world line, as opposed to the shortest path between the two? If so, maybe something like `shape` would be more fitting, as it could be easy to misconstrue this as de-normalized data.

AlmightyYakob:
> Additionally, DerivedRegion now has a reference to a VectorDataSource as its map representation, since it doesn't have a reference to a Dataset. We discussed how the region models themselves are internal representations, just like network nodes and edges. For visualization on the map, we should rely on DataSource -type objects. Therefore, DerivedRegions need a reference to one of these, which we know should be of the Vector type.

I think the problem with this is that derived regions can be "derived" from regions taken from multiple datasets. I've been thinking about the lingering gaps in our modeling,  and I'm starting to think we need an additional data source, `DerivedRegionDataSource`. This would replace `DerivedRegion` entirely, and would largely contain the same fields.

I also think that rather than point to a `dataset`, a `Region` should refer to a `VectorDataSource` (which in turn refers to a dataset), since I think that more closely models what we're trying to achieve with the `DataSource` abstraction. What do you think? I've included a diagram below to represent this configuration.

```mermaid
---
UVDAT ER Diagram
---
erDiagram
Dataset {
    CharField name
    TextField description
    ForeignKey city
}

DerivedRegionDataSource {
    CharField name
    ForeignKey city
    S3FileField geojson_data
    ManyToManyField source_regions
}

VectorDataSource {
    ForeignKey dataset
    JSONField metadata
    S3FileField geojson_data
}

Region {
    BigAutoField id
    CharField name
    JSONField properties
    MultiPolygonField boundary
    ForeignKey data_source
}
%% Relations

Dataset }|--|| City : city
VectorDataSource }|--|| Dataset: dataset
Region }|--|| VectorDataSource : data_source
DerivedRegionDataSource }|--|{ Region: source_regions
DerivedRegionDataSource }|--|| City: city
```

annehaley:
> I think the problem with this is that derived regions can be "derived" from regions taken from multiple datasets. I've been thinking about the lingering gaps in our modeling, and I'm starting to think we need an additional data source, DerivedRegionDataSource. This would replace DerivedRegion entirely, and would largely contain the same fields. I also think that rather than point to a dataset, a Region should refer to a VectorDataSource (which in turn refers to a dataset), since I think that more closely models what we're trying to achieve with the DataSource abstraction. 

I'm not sure we would be able to capture all we want to represent for the `DerivedRegion` in a `DataSource`. The goal for the `DataSource` type is to separate the map representation from the internal representation. For the map representation of a `DerivedRegion`, a `VectorDataSource` has everything we need. I think the internal representation of the boundary and the original regions should be separate from this. With the current model, it's fine if the original regions are from multiple datasets; the `DerivedRegion` is made independent from any datasets and a `VectorDataSource` (with `dataset=null`) can be made for it.

The way I think of the `DataSource` concept is as a implementation detail for visualization. Ideally, a user should be able to interact with every other model without concerning themselves with the `DataSource`s. They can fetch `Dataset`s and their `FileItem`s, or they can fetch `NetworkNode`s and `NetworkEdge`s related to a `Dataset`, or they can fetch `OriginalRegion`s and `DerivedRegion`s in a `City`. An advanced user may be interested in the converted data and may look at the `DataSource`s, but the average user would only want the main models. Thus, the `DataSource`-type models shouldn't contain any data essential to the objects themselves. They should be largely hidden from the user and consumed only by the map or chart viz.

uvdat/core/models/networks.py
annehaley: I changed it to `line_geometry` in [3e6b67f](https://github.com/OpenGeoscience/uvdat/pull/30/commits/3e6b67f84f49ae6c126218a9d3b3d907a9a81c98). Does this capture the purpose better?

uvdat/core/models/networks.py
```diff
@@ -0,0 +1,38 @@
...
+            .distinct()
+        )
+        to_edges = (
+            NetworkEdge.objects.filter(from_node=self.id)
+            .values_list('to_node_id', flat=True)
+            .distinct()
+        )
+        return NetworkNode.objects.exclude(id=self.id).filter(
+            models.Q(id__in=from_edges) | models.Q(id__in=to_edges)
+        )
```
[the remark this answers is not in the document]
annehaley: Updated this function in [3e6b67f](https://github.com/OpenGeoscience/uvdat/pull/30/commits/3e6b67f84f49ae6c126218a9d3b3d907a9a81c98)

uvdat/core/models/networks.py
annehaley: Ok I see what you mean. I changed the `related_name`s to "+" in [3e6b67f](https://github.com/OpenGeoscience/uvdat/pull/30/commits/3e6b67f84f49ae6c126218a9d3b3d907a9a81c98).

annehaley:
With the above consideration of how I think of `DataSource`s, I think it would be more appropriate to make `Chart`s as they were before. They should be a sibling to `Dataset`, not a child of `AbstractDataSource`. They aren't intended to be used by the map viz. I've made these changes in [1afb1f4](https://github.com/OpenGeoscience/uvdat/pull/30/commits/1afb1f4cddc9461594977d6ea956e9d262658852)

uvdat/core/models/networks.py
AlmightyYakob: I think so, thanks!

AlmightyYakob:
> The way I think of the DataSource concept is as a implementation detail for visualization. Ideally, a user should be able to interact with every other model without concerning themselves with the DataSources. They can fetch Datasets and their FileItems, or they can fetch NetworkNodes and NetworkEdges related to a Dataset, or they can fetch OriginalRegions and DerivedRegions in a City. An advanced user may be interested in the converted data and may look at the DataSources, but the average user would only want the main models. Thus, the DataSource-type models shouldn't contain any data essential to the objects themselves. They should be largely hidden from the user and consumed only by the map viz.

This makes sense to me, and resolves my questions/concerns with `DerivedRegion`. I guess for that specific use case, we would use the `city` field to retrieve the DerivedRegions, and when it came time to view them on the map, we would use the `VectorDataSource` to retrieve that data.

My only other concern then is making sure the original use case regarding time series is covered. I'm a bit fuzzy on that, maybe we should discuss this offline at some point.

annehaley:
Sounds good to me. I spent today on [bdcece6](https://github.com/OpenGeoscience/uvdat/pull/30/commits/bdcece69c8d3728bc8d6347252f3e0f437c027bf), which fits our data into these new models.  I tried combining the flood area datasets (grouping them by type) as a test for multiple `DataSource`s on the same `Dataset`. We would do the same thing with the time series data. 
Let me know what you think. We can discuss more tomorrow or next week if you'd like.

annehaley:
After adjusting the populate script to fit the latest design iteration, this is a visual representation of the objects that are created by the populate script, as well as their relationships. 
image

[384194e](https://github.com/OpenGeoscience/uvdat/pull/30/commits/384194e49dcdb990247bbeda37ef44f0d34c7ea0) Adds a txt file which shows the printed output from running the populate script.

AlmightyYakob:
That visual representation of our datasets makes things clear to me, thanks! I think I'm liking the way things look and feel, my only remaining question is how we plan to go about with the implementation? The changes involved are very large, would it make sense to have a long running base branch, that we would make PRs against? That way we're not reviewing the entire diff between master every time we want to make a change. I think we'll be iterating on the new design and functionality for a while.

annehaley:
I agree. I think this PR is going to be scoped to the models, rest, and populate script. Other changes, such as testing and web client updates, will go in separate PRs which are based on this branch. Let's still wait on merging this, though, until after those facets are complete, too (so we don't have a non-functional master for a while).

annehaley:
I've already made a PR for tests (#33) and I'm starting the web client updates today.

web/src/components/MainDrawerContents.vue
```diff
@@ -1,178 +1,140 @@
...
-    function fetchDatasets() {
-      getCityDatasets(currentCity.value.id).then((datasets) => {
-        currentCity.value.datasets = datasets;
-      });
+    // TODO: Avoid fetching entire map layer for each available
+    // derived region just to check if it's selected or not
+    async function derivedRegionSelected(derivedRegion: DerivedRegion) {
+      const mapLayer = await getMapLayerForDataObject(derivedRegion);
+      return isMapLayerVisible(mapLayer);
     }
```
AlmightyYakob, line 66: The only issue I've found when playing around with this branch is that the "Remove all" button (in the "Active Layers" menu) doesn't seem to work correctly when it comes to derived regions (the checkbox remains checked). I believe this function is at least partially responsible. If this can be fixed I think we're good to go.

annehaley: Ok, I changed this behavior in [dfe4c74](https://github.com/OpenGeoscience/uvdat/pull/30/commits/dfe4c7412974a057a4acfaac94a7a624ae80bffc) to mimic the structure we use for selected Datasets.

AlmightyYakob (approved):
I pushed [0a2c9ca](https://github.com/OpenGeoscience/uvdat/pull/30/commits/0a2c9ca1952dbb9e056a6b317e7b4d576bce03f5), which mostly just ensures that `newRegionName` is set back to `""` when a derived region selection is canceled (the rest is formatting/ergonomics). If that looks good to you, then this is ready.

PR #32, opened 2023-10-11
Fix Active Layers behavior

### LINKED ISSUE #31: Active Layers component behavior
The new component for "Active Layers" has some unexpected behavior. Sometimes labels are lost or repeated, and the drag-to-reorder behavior sometimes doesn't work. Sometimes the order of the labels does not represent the order of the layers. 

Tangential to this fix, I think it would be useful to have "X"s on each row to remove individual layers. Maybe the gear could be moved to the left after the grab icon to make room for an "X" on the right.

https://github.com/OpenGeoscience/uvdat/assets/44912689/3bfb6de3-dadc-4e89-8a85-61204a6a4e28

Also shown in this video: the labels in the "Available Datasets" are not wrapping properly, so they get cut off by the edge of the side drawer. Adding some CSS to make these wrap should fix this.

AlmightyYakob:
Fixes #31 
This PR also fixes the behavior in the raster slider / number input.

The underlying issue that was causing #31 to occur was the fact that we were deleting and re-adding a layer to the map, for two different use cases at the same time:

1. If we change any styling of a raster (because the style params are encoded in the layer source URL)
2. If we enable/disable the network vis mode of a network dataset (since they exist as two separate layers)

Since we heavily rely on consistent layer IDs, and removing/adding a layer to a map changes the layer ID, this was causing some weird behavior. This PR changes the behavior to directly handle raster styling changes, as well as node network visualization toggling.

### DIFF
--- a/web/src/components/MainDrawerContents.vue
+++ b/web/src/components/MainDrawerContents.vue
@@ -305,6 +305,9 @@ export default {
 .v-expansion-panel-text__wrapper {
   padding: 8px 10px 16px !important;
 }
+.v-checkbox {
+  max-width: 100%;
+}
 .expand-icon {
   float: right;
 }
--- a/web/src/components/OptionsDrawerContents.vue
+++ b/web/src/components/OptionsDrawerContents.vue
@@ -30,6 +30,10 @@ export default {
     const colormapRange = ref(undefined);
     const showConfirmConvert = ref(false);
 
+    // Use deep watcher to catch inputs from number fields alongside sliders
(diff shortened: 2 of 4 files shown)


annehaley (approved):
These changes LGTM, I tried out the behavior and didn't experience the problems I was having before. I did add one commit with a CSS rule, though, to fix the checkboxes for turning layers on/off. If that's ok with you, I believe this can be merged.

(P.S. Once this is merged, I'll make a copy of master to preserve this history as everything before the model redesign)

PR #33, opened 2023-10-19
Enable CI testing

annehaley:
Should be merged into #30 

Our CI was practically a noop before. This PR enables the tests and adds a pytest for the results of the populate script.
The `test` and `check-migrations` tests are run within the docker compose context, since our `django.Dockerfile` installs some large packages for `large-image`, which we need for some of our data ingest conversions during population (and to not raise import errors).


AlmightyYakob (approved):
Everything seems to work as far as I can tell

PR #34, opened 2023-10-23
Web updates

annehaley:
Depends on #33, which depends on #30. This is the last PR in the chain for the redesign effort. This PR is intended to reach feature parity with master, with one addition. This adds a slider to the options panel which allows the user to switch between map layers on a Dataset if there is more than one.


AlmightyYakob (approved):
Testing locally this seems to work as expected

PR #35, opened 2023-11-09
General Optimizations to new application design

AlmightyYakob:
Makes the following notable changes:

* Update the models to include new fields / indexes
* Update the behavior of vector map layer / tile retrieval, to make use of `tile_extents`, a field which contains the range of x/y values for every `z` depth in the `VectorMapLayer`
* Remove local paths for docker-compose service volumes (will need to re-run database migrations and population)
* Retrieve all map layers from one dataset endpoint, instead of two-staged approach


web/src/layers.ts
```diff
@@ -40,28 +69,14 @@ export async function getOrCreateLayerFromID(
     throw new Error(`Could not get or create openLayer for undefined layer`);
   }
 
-  let cachedMapLayer = _mapLayerManager.value.find((l) => {
+  const cachedMapLayer = _mapLayerManager.value.find((l) => {
     return l.id === mapLayerId && l.type === mapLayerType;
   });
   if (cachedMapLayer) return cachedMapLayer;
 
   const mapLayer = await getMapLayer(mapLayerId, mapLayerType);
```
annehaley, line 77: I believe with these changes we can get `mapLayer` (the full serialized version with the tile extents) from the dataset, so we wouldn't need this API call here anymore, and it will speed up the slider. But in this function, we would need to find it across all fetched dataset map layers, since the dataset (or derived region!) is not known here. Or we could refactor the function to take a dataset and index.

uvdat/core/tasks/dataset.py
```diff
@@ -36,18 +40,18 @@ def convert_dataset(
...
         for file_to_convert in FileItem.objects.filter(dataset=dataset):
             vector_map_layer = create_vector_map_layer(
                 file_to_convert,
                 style_options=style_options,
             )
-            if network_options:
+            if network:
                 create_network(vector_map_layer, network_options)
             elif region_options:
                 create_source_regions(vector_map_layer, region_options)
```
annehaley, line 57: From the CI failure, it looks like the `create_network` step isn't running. Perhaps the classification wasn't set.

annehaley: Changed in [1175a1b](https://github.com/OpenGeoscience/uvdat/pull/35/commits/1175a1b277492777e622e3381701ef0ce39c65c0)

web/src/layers.ts
annehaley: Changed in [3d4287f](https://github.com/OpenGeoscience/uvdat/pull/35/commits/3d4287f13ca924ae71c5339eee87ae87abe37d4d)

annehaley (approved):

PR #40, opened 2024-05-31
Publishing server code as Docker image

annehaley:
This PR changes the Dockerfile used by the Django and Celery containers to make it more portable. The same Dockerfile is now used to publish an image on `ghcr.io`, which is publicly accessible here: https://github.com/orgs/OpenGeoscience/packages?repo_name=uvdat. We can use this server image on other projects where only the client application needs custom code.

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: uvdat/uvdat:server
+  IMAGE_NAME: opengeoscience/uvdat:server
 
 jobs:
   publish-uvdat-server:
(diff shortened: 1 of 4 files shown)


johnkit (approved):
Good job getting the container image generated. In the future, I presume we'll add some kind of release/versioning process to distinguish between images generated by PRs and images generated from the main branch.

johnkit (approved):
Nice work. Thanks

annehaley:
@johnkit You made a good point about needing some versioning for the image tag. I added a Python Semantic Version step to the action, which will create a new release. We can use the release tag in the docker image tag. I also removed the clause that triggers this action on a pull request, so we will have to merge to test the behavior.

johnkit (approved):
+1

Though we've been discussing more architectural changes, let's go ahead and get this into CI

PR #43, opened 2024-06-24
Generate vector tiles dynamically from extracted features

jjnesbitt:
This PR replaces the use of pre-generated vector-tiles with pre-extracted vector features, that are then used for dynamic vector tiling. There is no caching layer included in this implementation (yet).


uvdat/core/admin.py
```diff
@@ -53,16 +52,6 @@ def get_dataset_name(self, obj):
         return obj.file_item.dataset.name
 
 
-class VectorTileAdmin(admin.ModelAdmin):
```
annehaley, line 56: Rather than just removing the `VectorTile` admin class, this should be replaced with an equivalent class for `VectorFeature` so that we can use the admin console to view and manage `VectorFeature` objects

uvdat/core/tasks/dataset.py
```diff
@@ -56,9 +56,5 @@ def convert_dataset(
             elif region_options:
                 create_source_regions(vector_map_layer, region_options)
 
-            # Create vector tiles after geojson_data may have
-            # been altered by create_network or create_source_regions
-            save_vector_tiles(vector_map_layer)
```
annehaley, line 61: After removing this line, is there still some part of the process that calls `save_vector_features` instead? Or should that new function be called here?

uvdat/core/tests/test_populate.py
```diff
@@ -43,4 +43,4 @@ def test_populate():
     assert SimulationResult.objects.all().count() == 0
     assert SourceRegion.objects.all().count() == 24
     assert VectorMapLayer.objects.all().count() == 5
-    assert VectorTile.objects.all().count() == 135
+    assert VectorFeature.objects.count() > 0
```
annehaley, line 46: Are you able to get a more exact assertion for the number of generated vector features?

uvdat/settings.py
```diff
@@ -52,6 +52,8 @@ def mutate_configuration(configuration: ComposedConfiguration) -> None:
             }
         }
 
+        configuration.CORS_ORIGIN_ALLOW_ALL = True
```
annehaley, line 55: Could you explain this addition?

jjnesbitt: Whoops, that's an accident. I'll remove it.

uvdat/core/tasks/dataset.py
jjnesbitt: Yes, it's called within `create_vector_map_layer`.

annehaley: The original purpose of this line was to save the vector data after the network/region interpretations have been completed. Both `create_network` and `create_source_regions` overwrite the geojson data on the map layer to include the IDs of the created nodes, edges, and source regions in the properties of the corresponding vector features. We should save vector features after this step to ensure those properties are propagated to the client (to be used for things like node deactivation, derived region creation)

web/src/types.ts
```diff
@@ -123,19 +123,11 @@ export interface RasterData {
...
-    [z: number]: {
-      min_x: number;
-      min_y: number;
-      max_x: number;
-      max_y: number;
-    };
-  };
-}
+// Just define type alias for now, since VectorMapLayer no longer extends any properties
+export type VectorMapLayer = AbstractMapLayer;
```
annehaley, line 127: Not sure if this change is the cause, but I get some type errors on this branch  when I start the client
image

uvdat/core/tasks/dataset.py
jjnesbitt: Fixed in [7748ede](https://github.com/OpenGeoscience/uvdat/pull/43/commits/7748edea88dc3c263aaabfe3c387e127ca3758a3)

web/src/types.ts
jjnesbitt: Seems like it was, due to the implications of type aliasing. It should be good now.

jjnesbitt: Actually not quite, let me fix the other errors.

jjnesbitt: Okay, good now.

annehaley (approved):

PR #45, opened 2024-07-11
Add New York Energy use case to populate script

annehaley:
- Add a new required argument `use_case` to the populate command. For the boston data, run `manage.py populate boston_floods`. For the new york data, run `manage.py populate new_york_energy`. Optional arguments `--include_large` and `--dataset_indexes` still work the same way.
- Reorganize `sample_data` folder to group use case data together
- Create custom `convert_dataset` functions for each use case
- Implement a module `nysdp.py` to pull vector features from `systemdataportal.nationalgrid.com`
- Interpret energy grid vector features as networks, export results to GeoJSONs, upload zipped results to DKC for download and import by others


annehaley:
After making modifications to speed up the algorithm which interprets network topology from the NYSDP basic features, running the algorithm across the whole state takes approximately 72 hours. I exported the results to geojsons (one for each county) and uploaded the data to DKC. The zipped results are approximately 200 MB. 

Next, I modified the populate script to download the results and save nodes and edges from that data. Running this script takes approximately 90 minutes. The output looks like this:
```
root@ebabfb6d27e4:/opt/uvdat-server# python manage.py populate new_york_energy --dataset_indexes 3 --include_large
Populating server with sample data for use case new_york_energy...
Creating Dataset objects...
        -  National Grid County Networks
         Converting data for National Grid County Networks...
                Creating network for Albany.
                Created 29517 nodes and 34042 edges.
                Creating network for Allegany.
                Created 2602 nodes and 2879 edges.
                Creating network for Cattaraugus.
                Created 9464 nodes and 10474 edges.
                Creating network for Cayuga.
                Created 330 nodes and 340 edges.
                Creating network for Chautauqua.
                Created 10107 nodes and 11398 edges.
                Creating network for Chenango.
                Created 63 nodes and 68 edges.
                Creating network for Clinton.
                Created 281 nodes and 310 edges.
                Creating network for Columbia.
                Created 6688 nodes and 7607 edges.
                Creating network for Hamilton.
                Created 1919 nodes and 2042 edges.
                Creating network for Herkimer.
                Created 6739 nodes and 7448 edges.
                Creating network for Jefferson.
                Created 15001 nodes and 17266 edges.
                Creating network for Rensselaer.
                Created 15432 nodes and 17358 edges.
                Creating network for Schoharie.
                Created 4892 nodes and 5343 edges.
                Creating network for St Lawrence.
                Created 12094 nodes and 13324 edges.
                Creating network for Warren.
                Created 10983 nodes and 12251 edges.
                Creating network for Washington.
                Created 8142 nodes and 8852 edges.
                Creating network for Wyoming.
                Created 1395 nodes and 1610 edges.
                Creating network for Cortland.
                Created 4011 nodes and 4464 edges.
                Creating network for Dutchess.
                Created 3 nodes and 2 edges.
                Creating network for Erie.
                Created 36508 nodes and 42876 edges.
                Creating network for Essex.
                Created 3564 nodes and 3886 edges.
                Creating network for Franklin.
                Created 5379 nodes and 5991 edges.
                Creating network for Fulton.
                Created 6062 nodes and 6665 edges.
                Creating network for Genesee.
                Created 7155 nodes and 8171 edges.
                Creating network for Lewis.
                Created 5391 nodes and 6160 edges.
                Creating network for Livingston.
                Created 4716 nodes and 5234 edges.
                Creating network for Madison.
                Created 5995 nodes and 6856 edges.
                Creating network for Monroe.
                Created 5533 nodes and 6492 edges.
                Creating network for Montgomery.
                Created 5864 nodes and 6589 edges.
                Creating network for Niagara.
                Created 14404 nodes and 16667 edges.
                Creating network for Oneida.
                Created 18852 nodes and 22381 edges.
                Creating network for Onondaga.
                Created 35068 nodes and 40780 edges.
                Creating network for Ontario.
                Created 1535 nodes and 1648 edges.
                Creating network for Orleans.
                Created 4649 nodes and 5200 edges.
                Creating network for Oswego.
                Created 12962 nodes and 14478 edges.
                Creating network for Otsego.
                Created 1475 nodes and 1598 edges.
                Creating network for Saratoga.
                Created 26271 nodes and 31447 edges.
                Creating network for Schenectady.
                Created 11819 nodes and 13374 edges.
        Completed in 5808.703211 seconds.
```

In total, 352,865  nodes and 403,571 edges are created.

There is still plenty of room for improvement on the network topology interpretation algorithm. There are still many nodes that act as connectors between only two edges where the line should have been merged, and there are many edges that run parallel to other edges (but this reflects the original features). The current algorithm relies on splitting up the state-wide network into multiple county networks, so inter-county connections are ignored. Speed would be another major area of improvement if this is ever revisited for future work.

sample_data/use_cases/new_york_energy/import_networks.py
```diff
@@ -0,0 +1,92 @@
...
+    NetworkEdge.objects.bulk_create(edges, batch_size=1000)
+    vector_features_from_network(network)
+    print(f'\t\tCreated {network.nodes.count()} nodes and {network.edges.count()} edges.')
+
+
+def perform_import(dataset, **kwargs):
+    print('\tEstimated time: 90 minutes.')
+    start = datetime.now()
+    Network.objects.filter(dataset=dataset).delete()
+    VectorMapLayer.objects.filter(dataset=network.dataset).delete()
```
BryonLewis, line 78: I don't think the scope of `network` is higher.  So where is it getting this from?  Should it just be `dataset`?

annehaley: Yes, thanks for catching that. I had pasted this line from somewhere else. Changed in [d9468e0](https://github.com/OpenGeoscience/uvdat/pull/45/commits/d9468e0c5427431f18ad9f2bbf387449b76f77a6).

sample_data/ingest_sample_data.py
```diff
@@ -46,9 +51,9 @@ def ingest_file(file_info, index=0, dataset=None, chart=None):
             new_file_item.file.save(file_path, ContentFile(f.read()))
 
 
-def ingest_contexts():
+def ingest_contexts(use_case):
     print('Creating Context objects...')
-    with open('sample_data/contexts.json') as contexts_json:
+    with open(str(USE_CASE_FOLDER / use_case / 'contexts.json')) as contexts_json:
```
jjnesbitt, line 56: I don't believe `str` is required here. Assuming that's the case, I'd recommend batching this change with the other two spots I suggested the change.
```suggestion
    with open(USE_CASE_FOLDER / use_case / 'contexts.json') as contexts_json:
```

sample_data/ingest_sample_data.py
```diff
@@ -66,9 +71,9 @@ def ingest_contexts():
             context_for_setting.datasets.set(Dataset.objects.filter(name__in=context['datasets']))
 
 
-def ingest_charts():
+def ingest_charts(use_case):
     print('Creating Chart objects...')
-    with open('sample_data/charts.json') as charts_json:
+    with open(str(USE_CASE_FOLDER / use_case / 'charts.json')) as charts_json:
```
jjnesbitt, line 76: Same here
```suggestion
    with open(USE_CASE_FOLDER / use_case / 'charts.json') as charts_json:
```

sample_data/ingest_sample_data.py
```diff
@@ -100,36 +105,24 @@ def ingest_charts():
             )
 
 
-def ingest_datasets(include_large=False, dataset_indexes=None):
+def ingest_datasets(use_case, include_large=False, dataset_indexes=None):
     print('Creating Dataset objects...')
-    with open('sample_data/datasets.json') as datasets_json:
+    with open(str(USE_CASE_FOLDER / use_case / 'datasets.json')) as datasets_json:
```
jjnesbitt, line 110: Same here
```suggestion
    with open(USE_CASE_FOLDER / use_case / 'datasets.json') as datasets_json:
```

sample_data/ingest_sample_data.py
```diff
@@ -143,12 +136,28 @@ def ingest_datasets(include_large=False, dataset_indexes=None):
...
+                        function = None
+                        try:
+                            module = importlib.import_module(module_path)
+                            function = getattr(module, function_name)
+                        except ImportError:
+                            print('\t', f'Failed to import module {module_path}.')
+                        except AttributeError:
+                            print('\t', f'Failed to import function {function_name} from {module_path}.')
+                        if function is not None:
+                            function(dataset_for_conversion, **func_kwargs)
```
jjnesbitt, line 153: I'd really like us to move away from defining modules and functions as static strings in JSON, and then later importing and using those things. It's very hard to understand what's happening from a readability perspective, and it breaks things like static type checking, import following, etc.

In this case the entire `ingest_datasets` script is being augmented to perform this module and function importing when it's only needed for the NYSDP data. I think we should define all of the individual use cases as python modules, with some standardized methods to call from the top level of those modules that correspond to what core feature is being created. You could do a lot with dataclasses, etc. as a way of defining the existing schema as Python objects.

For example, at the top level `__init__.py` of the `new_york_energy` module (and all other modules), you could define an `ingest_datasets` function, which would be imported by the master `ingest_datasets` script and ran. That way, each use case exposes the same "API surface" to the master ingest script, and any complexity specific to that use case is taken care of in that module.

I think this is important the continued development of UVDAT, as well as (and especially) all of the related projects. @annehaley let me know what you think about this and any questions/concerns you might have. I don't want to torpedo this PR, but there is also a lot of added code and changed features here. Maybe it makes sense to do this work as a very soon follow on PR, or even as a prerequisite PR that this would then re-base onto once merged.

jjnesbitt (changes requested):

sample_data/ingest_sample_data.py
annehaley: Changed in [3eda2f4](https://github.com/OpenGeoscience/uvdat/pull/45/commits/3eda2f4140862f1322782ad15956d7c9eac41493)

sample_data/ingest_sample_data.py
annehaley: Changed in [3eda2f4](https://github.com/OpenGeoscience/uvdat/pull/45/commits/3eda2f4140862f1322782ad15956d7c9eac41493)

sample_data/ingest_sample_data.py
annehaley: Changed in [3eda2f4](https://github.com/OpenGeoscience/uvdat/pull/45/commits/3eda2f4140862f1322782ad15956d7c9eac41493)

sample_data/ingest_sample_data.py
annehaley: @jjnesbitt Is this similar to what you had in mind: [5158311](https://github.com/OpenGeoscience/uvdat/pull/45/commits/5158311738d4d018f9f2d5db217e56ead8948975)? 

Most of the logic for both use case ingests is still shared, so that stayed at the top level. The only function that needed differentiation between the two use cases was what to do with the Dataset object once it is created, so each use case now has an `ingest` module that defines a function `convert_dataset`. 

For the Boston use case, this function spawns the `convert_dataset` task. For the New York use case, this function does the same unless the target dataset is "National Grid County Networks". The special behavior for that dataset is to use the custom import function. I also included a flag `PULL_LATEST` that is False by default. If this flag is set to True, then the full code for pulling new vector features from NYSDP and interpreting those features as county networks will run. I don't expect we'll do this often, but it at least ensures the code describes the procedure for that case.

sample_data/use_cases/new_york_energy/nysdp.py
```diff
@@ -0,0 +1,179 @@
...
+    print('\t\tDownloading basic features from NYSDP.')
+    feature_sets = None
+    with ThreadPoolExecutor(max_workers=len(include_services)) as pool:
+        feature_sets = pool.map(fetch_vector_features, include_services)
+
+    features = []
+    for feature_set in feature_sets:
+        for set_id, feature_set in feature_set.items():
+            for feature in feature_set:
+                properties = feature['properties']
```
BryonLewis, line 98: Just a small idea.  Noticed when working on the vector tile querying performance that the amount of data downloaded for the whole state network was quite high (~200MB).  Using ST_Simplify reduced that to about 100MB which still seemed a bit large.

Removing the Properties from the vector tiles got it down to ~12.5MB for all lines and points for the state.  Seeing as though most of the data on any particular node is empty could we remove it?  I want to make sure that for a lot of values it returns an empty string (`""`) and that doesn't mean something significantly different than not having the value.  I'm not sure if internally the vectortile PBF format condenses down the keys, but at least for UI purposes it would be nice to have all those empty keys not show up if they aren't in the data.

sample_data/ingest_sample_data.py
jjnesbitt: I think this is good for now. I may make a PR after the fact that attempts to further "standardize" things, since more on this topic is probably out of scope for this PR.

sample_data/use_cases/new_york_energy/nysdp.py
BryonLewis: Feel free to ignore this comment.  I may create a script that does this in EDNT separate from from the uploading the script.  That way I can reduce or not include empty properties.

Base you're merging on what Jacob says.  I don't have the full context of the UVDAT uses so I was only making comments on the process of loading the data into e-dnt.

BryonLewis:
I believe that the latest updates have broken the substation import (there is no call to the modules anymore).  I don't know if you knew this and were planning on further future modifications.

sample_data/use_cases/new_york_energy/nysdp.py
annehaley: Good idea; I think this is worth including in the import script to help reduce the size of the vector features in uvdat, too. It's a quick fix, so I made the change here: [e2c7549](https://github.com/OpenGeoscience/uvdat/pull/45/commits/e2c754949ff2ca66e7b67bdca4c9fb55f4aadbcb)

annehaley:
> ` [ee55020](/OpenGeoscience/uvdat/pull/45/commits/ee550201c20bd724d02a7d736a8807d88afb001b)`

@BryonLewis Thanks for pointing this out; I hadn't tested the refactoring thoroughly. I made changes in [ee55020](https://github.com/OpenGeoscience/uvdat/pull/45/commits/ee550201c20bd724d02a7d736a8807d88afb001b) to call the `create_vector_features` function manually since we won't use dynamic string imports.

uvdat/core/tasks/networks.py
```diff
@@ -213,6 +213,35 @@ def geojson_from_network(dataset):
     return new_geodata.to_json()
 
 
+def vector_features_from_network(network):
```
jjnesbitt, line 216: To clarify that this creates vector features and doesn't just return them:
```suggestion
def create_vector_features_from_network(network):
```

uvdat/core/tasks/networks.py
```diff
@@ -213,6 +213,35 @@ def geojson_from_network(dataset):
...
+    map_layer, created = VectorMapLayer.objects.get_or_create(dataset=network.dataset, index=0)
```
jjnesbitt, line 217:
```suggestion
    map_layer, _ = VectorMapLayer.objects.get_or_create(dataset=network.dataset, index=0)
```

jjnesbitt (changes requested):

uvdat/core/tasks/networks.py
annehaley: Changed in a287c78

uvdat/core/tasks/networks.py
annehaley: Changed in a287c78

jjnesbitt (approved):

PR #48, opened 2024-08-02
Change docker compose GH Action

annehaley:
Previously, we used a GH action from `adambirds` to run our CI tests in the docker compose environment. It appears that it stopped working recently, because we consistently get errors like this for both our `pytest` and `check-migrations` jobs:
image

This PR switches to a similar GH action from `hoverkraft-tech`. There is a bit of syntax difference, but it achieves the same goal of running our tests in the docker compose environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,6 @@ on:
       - master
 
 jobs:
-
   lint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


jjnesbitt (approved):
I'd say eventually we should run our tests and services directly on github actions, instead of docker compose, but there's some downsides to that as well and this is a positive incremental improvement.

PR #60, opened 2024-08-15
Authentication and projects

### LINKED ISSUE #52: Implement user authentication

### LINKED ISSUE #53: Replace Context model with Project model

annehaley:
Resolves #52. Resolves #53.

This PR adds users and authentication to UVDAT. With this change, we can also change the structure with which we group Datasets. Instead of Contexts, we now have Projects, which have all the functionality of Contexts plus user access control. A user can belong in one of three categories on a given Project: 1) followers have read-only permissions, 2) collaborators have read-write permissions, and 3) the owner has read, write, and delete permissions. Superusers have read, write, and delete permissions on everything. 

UI changes to provide an interface for Project configuration are in #66.


annehaley:
@jjnesbitt I discovered that the OpenLayers implementation of tile loadiing will not work with this branch due to a lack of authentication. I spent a while trying to add authentication to those requests, but I didn't want to expend too much effort getting it to work, since you'll be writing a MapLibre implementation. Let's hold off on this branch until after switching to MapLibre. I'll base my branch on yours and add authentication to that implementation.

faiza-a:
[Figma Mockups link.](https://www.figma.com/design/OQ7RqVWPSlsyNMcLeI4j69/UVDAT?node-id=0-1&node-type=CANVAS&t=GJoibRpxC04Iu9LQ-0)

uvdat/core/rest/filter.py
```diff
@@ -0,0 +1,21 @@
...
+        if request.method == 'GET':
+            valid_ids = [o.id for o in queryset if o.readable_by(user) and o.id in valid_ids]
+        elif request.method == 'PUT' or request.method == 'PATCH':
+            valid_ids = [o.id for o in queryset if o.editable_by(user) and o.id in valid_ids]
+        elif request.method == 'DELETE':
+            valid_ids = [o.id for o in queryset if o.deletable_by(user) and o.id in valid_ids]
+        elif request.method == 'POST':
+            # no access control required for POST requests
+            pass
+        return queryset.filter(id__in=valid_ids)
```
jjnesbitt, line 21: There's a couple things going on here that I'd like to address.

---

The first is that this DRF filter is doing two things at once in a non-standard way. It's _filtering_ the objects visible to the user not only based on what the user has read access to, but based on _type_ of request made. This means that if a user who only has read access to a project, for example, tries to make a `PUT` request on a resource that they are able to view, they'll be returned a `404 Not Found`. This should instead be handled in two separate steps:

1. Filtering objects for which access to the user is not granted. An example of this is if a user lists projects, any project that they don't have access to will not appear in the listing (will be filtered out). If the user attempts to make _any_ request on a project to which they don't have access to, they'll be returned a `404 Not Found`, because in reference to the list of objects the user has access to, it doesn't exist.
2. Enforce the hierarchy of permissions that pertains to the request. So, for example, if a user had "follower" access to a project, and tried to make a `PUT` request, it would return `403 Forbidden`, because while the user can _read_ the object in question, they do not have further permissions. 

Regarding object permissions, this should probably be handled using [DRF permission classes](https://www.django-rest-framework.org/api-guide/permissions/#permissions), as they exist to handle this exact use case. [Here](https://github.com/dandi/dandi-archive/blob/99e55f419a9a68f5f382c2645d367bb43d8e26ad/dandiapi/api/permissions.py#L20) is an example of such use on the DANDI project.

---

The second thing is that I think the cascading of the permissions methods (e.g. `readable_by`) to essentially all models in the codebase, is an anti-pattern. Not only is it confusing and cumbersome, but it's prone to error as the same methods are being defined in every model class, just to call the identical method in their "parent relation" class, which eventually reaches a project. For example, the string `readable_by` appears 21 times in the codebase, and only 1 of those is the actual use of the method in the filter. This is a huge maintenance burden imo.

In reality, since there's only one level at which we enforce these permissions (the project level), there's no need to have methods pertaining to permissions on each model class. This can be moved to each models ViewSet, or may even be able to be consolidated into one DRF Permission class, that can then be invoked in each ViewSet via the `permission_classes` field.

---

Lmk what you think, I'd be happy to discuss this in more depth.

uvdat/core/rest/regions.py
```diff
@@ -16,14 +17,18 @@
...
-class SourceRegionViewSet(mixins.RetrieveModelMixin, mixins.ListModelMixin, GenericViewSet):
+class SourceRegionViewSet(ModelViewSet):
     queryset = SourceRegion.objects.all()
     serializer_class = SourceRegionSerializer
+    filter_backends = [AccessControl]
+    lookup_field = 'id'
 
 
-class DerivedRegionViewSet(mixins.RetrieveModelMixin, mixins.ListModelMixin, GenericViewSet):
+class DerivedRegionViewSet(ModelViewSet):
```
jjnesbitt, line 27: What's the reason for replacing these mixins with `ModelViewSet`? This exposes extra unused API endpoints.

uvdat/urls.py
```diff
@@ -49,6 +52,7 @@
     path('api/v1/', include(router.urls)),
     path('api/docs/redoc/', schema_view.with_ui('redoc'), name='docs-redoc'),
     path('api/docs/swagger/', schema_view.with_ui('swagger'), name='docs-swagger'),
+    path('', RedirectView.as_view(url=settings.HOMEPAGE_REDIRECT_URL)),  # type: ignore
```
jjnesbitt, line 55:
```suggestion
    # Re-direct any requests to the base application, back to the vue app
    path('', RedirectView.as_view(url=settings.HOMEPAGE_REDIRECT_URL)),  # type: ignore
```

web/src/components/map/Map.vue
```diff
@@ -142,13 +159,24 @@ export default {
...
+    <div class="base-layer-control">
+      <v-btn icon>
+        <v-checkbox
+          v-model="showMapBaseLayer"
+          true-icon="mdi-map-check"
+          false-icon="mdi-map-outline"
+          style="margin-top: -4px"
+        />
+      </v-btn>
+    </div>
```
jjnesbitt, line 179: This doesn't seem to actually disable the base map layer.

web/src/components/map/Map.vue
```diff
@@ -55,6 +62,16 @@ export default {
...
+        transformRequest: (url) => {
+          let headers = {};
+          if (!url.includes("openstreetmap")) {
+            headers = oauthClient?.authHeaders;
+          }
+          return {
+            url,
+            headers,
+          };
+        },
```
jjnesbitt, line 74: While I think I understand what's going on here (only passing auth headers to our API), I think a comment explaining that would be helpful.

web/src/App.vue
```diff
@@ -27,63 +26,111 @@ export default defineComponent({
...
+          style="float: right"
+        >
+          <v-icon>mdi-close</v-icon>
+        </v-btn>
+        <v-card-title> Error: </v-card-title>
+        <v-card-text>
+          {{ currentError }}
+        </v-card-text>
+      </v-card>
     </v-overlay>
```
jjnesbitt, line 98: Was there an underlying reason this was added? Is there an error that's hit fairly often? Or is it just a catch-all?

jjnesbitt (changes requested):

web/src/App.vue
annehaley: This is a catch-all; previously, the error was shown in the middle of the toolbar. At one point, I got a very long error text that messed with the alignment of the logout button on the right side of the toolbar, so I decided to move the error display to a dialog.

web/src/components/map/Map.vue
annehaley: Whoops, this is due to not organizing commits between my two PRs properly. There is a commit in the other PR that adds the watcher on `showMapBaseLayer`. I'll move that change to this PR.

uvdat/core/rest/regions.py
annehaley: This change was to be congruent with other models and offer a more complete API, but I'm open to reverting this if you think that we shouldn't. The Regions will be reworked soon, anyway.

uvdat/settings.py
```diff
@@ -20,6 +21,11 @@ class UvdatMixin(ConfigMixin):
 
     BASE_DIR = Path(__file__).resolve(strict=True).parent.parent
 
+    # Override default signup sheet to ask new users for first and last name
+    ACCOUNT_FORMS = {'signup': 'uvdat.core.rest.accounts.AccountSignupForm'}
```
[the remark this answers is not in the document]
annehaley: The Account signup page should look like this: 
image

It should include fields for first and last name. It inherits from the allauth signup form we use by default in Resonant, but it adds these fields. The `ACCOUNT_FORMS` setting is documented by `django-allauth` here: https://docs.allauth.org/en/latest/account/configuration.html#:~:text=their%20email%20address.-,ACCOUNT_FORMS,-Used%20to%20override

Let me know if you don't see the inputs for first and last name on your signup page.

uvdat/core/models/project.py
```diff
@@ -0,0 +1,29 @@
...
+    collaborators = models.ManyToManyField(User, related_name='read_write_projects')
+    followers = models.ManyToManyField(User, related_name='read_only_projects')
+
+    def readable_by(self, user):
+        return (
+            user.is_superuser
+            or user == self.owner
+            or user in self.collaborators.all()
+            or user in self.followers.all()
+        )
```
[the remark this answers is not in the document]
annehaley: I see what you mean. Another library could manage permissions for us and eliminate some of these extra model methods. `django-guardian` is great, but it doesn't have any support for permission inheritance. In order to cascade permissions from the `Project` to its associated objects, there would be a lot of overriding behavior on each Model (or worse, manually updating individual permissions on associated objects).

But in doing this research I found `django-improved-permissions`, which does support permission inheritance and would allow us to define the Owner, Collaborator, and Follower roles explicitly. Any objections to trying this library instead? https://django-improved-permissions.readthedocs.io/en/latest/inheritance.html

sample_data/ingest_use_case.py
```diff
@@ -54,26 +54,26 @@ def ingest_file(file_info, index=0, dataset=None, chart=None):
...
+                    project_for_setting = Project.objects.create(
+                        name=project['name'],
+                        default_map_center=Point(*project['default_map_center']),
+                        default_map_zoom=project['default_map_zoom'],
                     )
-                    print('\t', f'Context {context_for_setting.name} created.')
+                    print('\t', f'Project {project_for_setting.name} created.')
 
-                context_for_setting.datasets.set(Dataset.objects.filter(name__in=context['datasets']))
+                project_for_setting.datasets.set(Dataset.objects.filter(name__in=project['datasets']))
```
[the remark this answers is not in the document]
annehaley: Changed in [318f695](https://github.com/OpenGeoscience/uvdat/pull/60/commits/318f695abe7b3652fff3bbba4804b2113ce8c256)

uvdat/core/management/commands/makeclient.py
```diff
@@ -0,0 +1,45 @@
...
+                client_type='public',
+                authorization_grant_type='authorization-code',
+                skip_authorization=True,
+            )
+            application.save()
+            self.stdout.write(self.style.SUCCESS('Client Application created.'))
+        except User.DoesNotExist:
+            raise CommandError(
+                'A user must exist before creating a client. Use createsuperuser command.'
+            )
```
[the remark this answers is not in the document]
annehaley: Changed in [88af424](https://github.com/OpenGeoscience/uvdat/pull/60/commits/88af42465644d78751e9859e605beb351d9a5f3c)

uvdat/core/models/dataset.py
```diff
@@ -16,11 +16,20 @@ class DatasetType(models.TextChoices):
...
 
-    def is_in_context(self, context_id):
-        from uvdat.core.models import Context
+    def is_in_project(self, project_id):
+        from uvdat.core.models import Project
 
-        context = Context.objects.get(id=context_id)
-        return context.datasets.filter(id=self.id).exists()
+        project = Project.objects.get(id=project_id)
+        return project.datasets.filter(id=self.id).exists()
```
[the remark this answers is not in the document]
annehaley: Changed in [d78a494](https://github.com/OpenGeoscience/uvdat/pull/60/commits/d78a4941266da8f0a5ca541a512b3c83cc7bb2ef)

web/src/components/map/Map.vue
annehaley: Added back in [df03bad](https://github.com/OpenGeoscience/uvdat/pull/60/commits/df03badcd8e4c15c16bb0d23571351ba517fcf49)

web/src/components/map/Map.vue
annehaley: Comment added in [9f6c9ab](https://github.com/OpenGeoscience/uvdat/pull/60/commits/9f6c9ab78e1e8784a4c1a2530be09c0b60dfc666)

uvdat/urls.py
annehaley: Comment added in [af567f1](https://github.com/OpenGeoscience/uvdat/pull/60/commits/af567f19997a7b4d95f8560f88aac64a5f904776)

uvdat/core/models/project.py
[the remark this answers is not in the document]
jjnesbitt: > `django-guardian` is great, but it doesn't seem to have any support for permission inheritance. In order to cascade permissions from the `Project` to its associated objects, there would be a lot of overriding behavior on each Model (or worse, manually updating individual permissions on associated objects).

I don't understand why you'd need to cascade any permissions from the project, to all of it's associated objects. We only need to give a user permission on the Project, since that reflects our permission model. If you need to check if a user has access to another object contained with a project, like a network, all you have to do semantically is ask "does this user have the required permissions on this network's project?". There's no cascading required. In django-guardian, this check would look something like this:

```python
user.has_perm(['owner', 'collaborator', 'follower'], network.dataset.project_set)
```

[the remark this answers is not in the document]
annehaley: I was hoping to cascade permissions so that we don't have to differ what we send to `has_perm` for each model, since the relationship path will differ depending on the type of object we receive. Implementing this in a DRF permission class like you suggest in another comment, how can we receive an object of an unknown type and call `user.has_perm(perms, obj)` where if `obj` is a Chart, we check perms on `obj.project` and if `obj` is a Project, we check perms on `obj` directly? I've been searching the docs for some model function to override. If this isn't possible, we may have a bunch of `if` statements in our permission class, and I was hoping to avoid putting model logic in the permission class.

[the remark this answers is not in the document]
annehaley: One option could be to implement a `has_permission` method on each model. The one for Project would just be a wrapper around guardian's `has_perm`, and all other models would have a custom implementation that leverages project associations. The DRF permission class could just call these instead.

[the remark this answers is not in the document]
jjnesbitt: It seems to me we have two options:

1. Define a method in each model class or viewset that returns if the user has the required permission on the object. This would be similar to the existing approach, and I'd prefer to avoid this for the same reasons. It spreads out the permission logic across the whole codebase, and is cumbersome.
2. Define a permission class that is used in all/most views, that uses different queries for different models. This would involve if statements (or more ideally a [`match`](https://docs.python.org/3/tutorial/controlflow.html#match-statements) statement), but I don't see the problem with that. It would consolidate the permission logic in one place.

I much prefer the second option. Since we'd probably handle list requests through filtering, the DRF permission class `has_object_permission` method could be the main spot where this is defined, and that passes in a model object. In that case, the logic could be fairly trivial. Something like the following:

```python
def get_obj_queryset(obj):
    if isinstance(obj, Network) or isinstance(obj, FileItem):
        return obj.dataset.project_set

    if isinstance(obj, Chart):
        return obj.project
    ...


class UVDATPermission(BasePermission):
    def has_object_permission(self, request, view, obj) -> bool:
        if request.method in SAFE_METHODS:
            return True

        queryset = get_obj_queryset(obj)
        return request.user.has_perm(['owner', 'collaborator', 'follower'], queryset)
```

jjnesbitt:
Looking at this again, I'm a bit confused on the permissions model. Datasets are supposed to be readable by everyone, and modifiable only by superusers, correct? And most models refer (directly or indirectly) to a dataset. Which means that most models are readable by everyone and only modifiable by superusers? The exceptions being `Chart` and `AnalysisResult`?

If this is case, I think we're getting way ahead of ourselves, since all we _really_ need to do is implement permissions for charts and analysis results. The permission model of "anyone can read, only superusers can edit" isn't really a permission model, it's what you'd get before adding permissions. I think we should save ourselves a lot of time and effort, and simply remove any `PUT`, `POST` or `PATCH` endpoints on any model that can only be modified by a superuser. If there's no way for a normal user to make use of these endpoints, there's no reason to have them exposed.

annehaley:
We previously discussed adding permissions to Datasets as well, with a `private` bool field and a similar concept of ownership. These permissions will cascade to the objects related (directly or indirectly) to a dataset.  We still have to discuss the details as a team, but we decided to tackle this in a separate PR and milestone effort. The structure we set up in this PR for Project-related items will be used in a similar manner for Dataset-related items.

uvdat/core/rest/guardian.py
```diff
@@ -0,0 +1,51 @@
...
+from rest_framework.permissions import SAFE_METHODS, IsAuthenticated
+
+from uvdat.core import models
+
+
+class GuardianPermission(IsAuthenticated):
+    def get_object_queryset(self, obj):
+        if isinstance(obj, models.Project):
+            return obj
+        elif isinstance(models.Dataset):
```
jjnesbitt, line 12: Just a general code style suggestion, the "else" part of the `elif` is never reached:
```suggestion
        if isinstance(obj, models.Project):
            return obj
        if isinstance(models.Dataset):
```

uvdat/core/rest/guardian.py
```diff
@@ -0,0 +1,51 @@
...
+        ):
+            return obj.project
+        elif (
+            isinstance(obj, models.FileItem)
+            or isinstance(obj, models.VectorMapLayer)
+            or isinstance(obj, models.RasterMapLayer)
+            or isinstance(obj, models.Network)
+            or isinstance(obj, models.SourceRegion)
+        ):
+            return obj.dataset.project_set
```
jjnesbitt, line 27: I just learned this neat fact about `isinstance`, from the [docs](https://docs.python.org/2/library/functions.html#isinstance):

> If classinfo is a tuple of class or type objects (or recursively, other such tuples), return true if object is an instance of any of the classes or types

```suggestion
        if isinstance(
            obj,
            (
                models.FileItem,
                models.VectorMapLayer,
                models.RasterMapLayer,
                models.Network,
                models.SourceRegion,
            ),
        ):
            return obj.dataset.project_set
```

uvdat/core/rest/guardian.py
```diff
@@ -0,0 +1,51 @@
...
+    def has_object_permission(self, request, view, obj):
+        if request.user.is_superuser:
+            return True
+
+        perms = ['follower']
+        if request.method not in SAFE_METHODS:
+            perms.append('collaborator')
+        if request.method == 'DELETE':
+            perms.append('owner')
+        return request.user.has_perm(perms, self.get_object_queryset(obj))
```
jjnesbitt, line 40: Since we only assign one permission per user <-> project pairing, we just need to check for the lowest single permission required. Checking for all three would always fail, I think:

```suggestion
        perm = 'follower'
        if request.method == 'DELETE':
            perm = 'owner'
        elif request.method not in SAFE_METHODS:
            perm = 'collaborator'

        return request.user.has_perm(perm, self.get_object_queryset(obj))
```

uvdat/core/rest/guardian.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class GuardianFilter(BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        if request.user.is_superuser:
+            pass
+        return get_objects_for_user(
+            klass=queryset,
+            user=request.user,
+            perms=['follower', 'collaborator', 'owner'],
```
jjnesbitt, line 50: We need the `any_perm` parameter here, or it will require all 3 perms be present at once.
```suggestion
            perms=['follower', 'collaborator', 'owner'],
            any_perm=True,
```

jjnesbitt:
> We previously discussed adding permissions to Datasets as well, with a `private` bool field and a similar concept of ownership. These permissions will cascade to the objects related (directly or indirectly) to a dataset. We still have to discuss the details as a team, but we decided to tackle this in a separate PR and milestone effort. The structure we set up in this PR for Project-related items will be used in a similar manner for Dataset-related items.

That all makes sense, but I don't see how that warrants exposing all available methods of all model endpoints, when the majority of them are unused. It just creates clutter and makes it harder to reason about the structure and uses of the API, as well as a potential violation of our permission model, if not carefully managed. Since we're constructing a new part of the codebase, we should avoid making assumptions about how we're going to be doing things in the future. It's easier to build more or slightly tweak things going forward than it is to tear down everything and rebuild it.

Backing out any unused endpoints is probably out of scope for this PR, so I may make a follow up PR once this is merged. To loop back to [my above comment](https://github.com/OpenGeoscience/uvdat/pull/60#discussion_r1777749318), I think we should refrain from exposing unused endpoints, just for the sake of congruence.

uvdat/core/rest/guardian.py
annehaley: Added in [f831cf3](https://github.com/OpenGeoscience/uvdat/pull/60/commits/f831cf38eefd71fce0d36d418bb3e6810b6ef144)

uvdat/core/rest/guardian.py
annehaley: Updated in [f831cf3](https://github.com/OpenGeoscience/uvdat/pull/60/commits/f831cf38eefd71fce0d36d418bb3e6810b6ef144)

uvdat/core/rest/guardian.py
annehaley: Changed in [f831cf3](https://github.com/OpenGeoscience/uvdat/pull/60/commits/f831cf38eefd71fce0d36d418bb3e6810b6ef144)

uvdat/core/rest/regions.py
annehaley: Change undone in [e317c6d](https://github.com/OpenGeoscience/uvdat/pull/60/commits/e317c6d3e327500324eb28f6d1df1c3656620c74)

uvdat/core/models/project.py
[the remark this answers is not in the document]
annehaley: `django-guardian` added in b40dd60. All permissions logic is consolidated to `uvdat/core/rest/guardian.py`.

uvdat/core/rest/filter.py
annehaley: Changed in [b40dd60](https://github.com/OpenGeoscience/uvdat/pull/60/commits/b40dd6078ddd1e3341d75dcb2a502379753c2316). Access control implementation was split between a permission class for detail views and a filter backend for list views. `readable_by` and other permissions methods were removed from models.

uvdat/core/rest/guardian.py
```diff
@@ -0,0 +1,99 @@
...
+            perms = ['owner']
+        allowed_objects = filter_by_parent_perms(obj.__class__, request.user, perms=perms, obj=obj)
+        return allowed_objects.filter(id=obj.id).exists()
+
+
+class GuardianFilter(BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        if request.user.is_superuser:
+            return queryset
+        return filter_by_parent_perms(queryset.model, request.user, queryset=queryset)
```
jjnesbitt, line 99: `filter_by_parent_perms` is much more complicated than it needs to be, since we can just leverage queryset filters to accomplish what we want. Due to the fact that there are only 4 "groups" of queryset filters (4 different syntaxes needed to filter projects from the base model), it becomes even easier to read. This code passes the tests you added.

```suggestion
# TODO: Dataset permissions should be separated from Project permissions
def filter_queryset_by_project_permission(
    queryset: QuerySet[Model], user: User, perms: list[str] | None = None
):
    if perms is None:
        perms = ['follower', 'collaborator', 'owner']

    # Get all projects a user has access to
    user_projects = get_objects_for_user(
        klass=models.Project, user=user, perms=perms, any_perm=True
    )

    model = queryset.model
    if model == models.Project:
        return queryset.filter(id__in=user_projects.values_list('id', flat=True))
    if model in [models.Dataset, models.Chart, models.DerivedRegion]:
        return queryset.filter(project__in=user_projects)
    if model in [
        models.FileItem,
        models.RasterMapLayer,
        models.VectorMapLayer,
        models.Network,
        models.SourceRegion,
    ]:
        return queryset.filter(dataset__project__in=user_projects)

    if model in [models.NetworkNode, models.NetworkEdge]:
        return queryset.filter(network__dataset__project__in=user_projects)

    # If any models are un-caught, raise an exception
    raise NotImplementedError


class GuardianPermission(IsAuthenticated):
    def has_object_permission(self, request, view, obj):
        if request.user.is_superuser:
            return True

        perms = ['follower', 'collaborator', 'owner']
        if request.method not in SAFE_METHODS:
            perms = ['collaborator', 'owner']
        if request.method == 'DELETE':
            perms = ['owner']

        if not isinstance(obj, Model):
            raise Exception('Only Django models may be used in permission check')

        # Create queryset out of single object, so it can be passed to the filter function
        queryset = obj.__class__.objects.filter(pk=obj.pk)

        # If the object remains in the queryset after this function filters it, then the user has
        # the required permission on at least one associated project
        return filter_queryset_by_project_permission(queryset, request.user, perms).exists()


class GuardianFilter(BaseFilterBackend):
    def filter_queryset(self, request, queryset, view):
        if request.user.is_superuser:
            return queryset

        return filter_queryset_by_project_permission(queryset, request.user)
```

annehaley: Thanks! Changed in [9989e6f](https://github.com/OpenGeoscience/uvdat/pull/60/commits/9989e6f6aee6aaf3377cec43907911476c1e05e5)

uvdat/core/rest/guardian.py
annehaley: Outdated; logic rewritten in [9989e6f](https://github.com/OpenGeoscience/uvdat/pull/60/commits/9989e6f6aee6aaf3377cec43907911476c1e05e5)

PR #70, opened 2024-09-25
Implement GCC algorithm in native Postgres

jjnesbitt:
Since we're dealing with larger networks now (i.e. the boston roads dataset), the current algorithm for finding the GCC doesn't perform well enough. An example query on the boston roads dataset takes ~46 seconds. That same query takes ~40ms with this native Postgres query.

@annehaley PTAL and let me know if this maintains the same semantics as the original query.

### DIFF
--- a/uvdat/core/models/networks.py
+++ b/uvdat/core/models/networks.py
@@ -1,8 +1,45 @@
 from django.contrib.gis.db import models as geo_models
-from django.db import models
-import networkx as nx
+from django.db import connection, models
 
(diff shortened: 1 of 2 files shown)


annehaley (commented):
Impressive time delta. The original implementation was done with `networkx`, so I'm not sure about the semantics of the previous algorithm. This new method seems reasonable. Could you either remove the old `get_gcc` method from the `Network` model or move this logic there?

annehaley (approved):

PR #78, opened 2024-10-01
Prevent writing of file if request was not successful

### LINKED ISSUE #72: Fix flaky pytest tests
The [`test_populate`](https://github.com/OpenGeoscience/uvdat/blob/82dd93c0f4f15ea155c02a75d0d6152d180b6b53/uvdat/core/tests/test_populate.py#L23) test downloads external data as a part of the test. This means that if the download fails for any reason (very common with web requests, especially if the host is unreliable), the test will fail. This isn't good practice, as it means that tests can pass or fail independent of any changed code in our codebase. We should aim for our tests to be reproducible, that is, given the same commit in our code base, the test should produce the same result, every time.

To allow for our tests to be reproducible, we can employ the following strategies:

1. Test at the appropriate level. For example, the `populate` management script really just wraps `ingest_use_case`. We should just test that function directly, to allow for more robust tests, instead of relying on proxy functions like `call_command`, which we don't have much control over.
2. Mock data. Rather than relying on the downloading of datasets in our tests, we can mock or generate the minimum data we need to accomplish the task. 
3. Limit test scope. In the above example, `ingest_use_case` really just calls `ingest_datasets`, `ingest_projects`, and `ingest_charts`, sequentially and in that order. We should realistically just test those functions directly, which will give us more control, and isolate the setup to each individual use case. If we then want to test `ingest_use_case` overall, we don't need to test every part of all the subsequent functions, since those are already covered by the individual tests. We can minimally just test that all 3 of them ran successfully.

jjnesbitt:
Closes #72

Currently, the request to retrieve certain files from DKC often fails with a `502 Bad Gateway`, but the file is still written to anyway. Since the file is only written to the first time, this means that if this bad request happens when first running `populate`, the bad file will remain there until the file item is deleted.

### DIFF
--- a/sample_data/ingest_use_case.py
+++ b/sample_data/ingest_use_case.py
@@ -32,6 +32,7 @@ def ingest_file(file_info, index=0, dataset=None, chart=None):
         file_location.parent.mkdir(parents=True, exist_ok=True)
         with open(file_location, 'wb') as f:
             r = requests.get(file_url)
+            r.raise_for_status()
             f.write(r.content)
 
     existing = FileItem.objects.filter(name=file_name)


annehaley (approved):
I agree we should add this line, but I'm not sure it closes the issue. Tests that encounter the nginx 502 error will still fail, albeit with a more informative exception raised.

jjnesbitt:
> I agree we should add this line, but I'm not sure it closes the issue. Tests that encounter the nginx 502 error will still fail, albeit with a more informative exception raised.

You're right, it more just fixes the issue for local dev, since you can just re-run it until it succeeds. I'll remove the link to that issue so this PR won't close it, and reword the issue a bit.

PR #80, opened 2024-10-09
Version Tags

### LINKED ISSUE #65: Add semantic version label to UI
@aashish24 requested that we add a version label to the UI. This could go in the toolbar near "UVDAT" title (either as a tooltip or subtitle) or we could add a footer to the bottom of the page. Add a button to copy version tag, date, and hash to clipboard.

Also, we should update the version tag in `setup.py` and `package.json`.

annehaley:
This PR adds a menu to the toolbar title, which displays the following on hover:

- A link to this repository
- The current version string (full string if the repo is "dirty", tag-only otherwise) and a button to copy to clipboard
- The current version hash and a button to copy to clipboard

This PR also configures semantic release to update the version strings in `setup.py` and `web/package.json` upon release.

Resolves #65.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,8 @@ from pathlib import Path
 
 from setuptools import find_packages, setup
 
+__version__ = '0.4.1'
+
 readme_file = Path(__file__).parent / 'README.md'
 if readme_file.exists():
     with readme_file.open() as f:
@@ -12,7 +14,7 @@ else:
 
 setup(
     name='uvdat',
-    version='0.1.0',
+    version=__version__,
     description='',
     long_description=long_description,
     long_description_content_type='text/markdown',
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -2,6 +2,8 @@
 
 from setuptools import find_packages, setup
 
+__version__ = '0.4.1'
```
jjnesbitt, line 5: Do you know if this will be automatically updated by semantic release?

annehaley: According to [the documentation](https://python-semantic-release.readthedocs.io/en/latest/configuration.html#version-variables), setting `version_variables` in our `pyproject.toml` configures semantic release to update this string (and the version string in `package.json`). However, I am not able to test this without making a release.

PR #83, opened 2024-10-11
File Deletion Signals

### LINKED ISSUE #79: Proper MinIO/S3 file deletion.
The data isn't getting properly removed from MinIO/S3 when deleting datasets/file-items/map layers.

https://github.com/OpenGeoscience/uvdat/blob/master/uvdat/core/models/file_item.py#L15

```
@receiver(models.signals.pre_delete, sender=FileItem)
def delete_content(sender, instance, **kwargs):
    if instance.file:
        instance.file.delete(save=False)
```

https://github.com/OpenGeoscience/uvdat/blob/master/uvdat/core/models/map_layers.py#L26
```
@receiver(models.signals.pre_delete, sender=RasterMapLayer)
def delete_raster_content(sender, instance, **kwargs):
    if instance.cloud_optimized_geotiff:
        instance.cloud_optimized_geotiff.delete(save=False)
```
https://github.com/OpenGeoscience/uvdat/blob/master/uvdat/core/models/map_layers.py#L44
```
@receiver(models.signals.pre_delete, sender=VectorMapLayer)
def delete__vectorcontent(sender, instance, **kwargs):
    if instance.geojson_file:
        instance.geojson_file.delete(save=False)
```

annehaley:
Resolves #79.

### DIFF
--- a/uvdat/core/models/file_item.py
+++ b/uvdat/core/models/file_item.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.dispatch import receiver
 from django_extensions.db.models import TimeStampedModel
 from s3_file_field import S3FileField
 
@@ -21,3 +22,9 @@ class FileItem(TimeStampedModel):
     def download(self):
         # TODO: download
         pass
+
+
+@receiver(models.signals.pre_delete, sender=FileItem)
+def delete_content(sender, instance, **kwargs):
+    if instance.file:
+        instance.file.delete(save=False)
--- a/uvdat/core/models/file_item.py
+++ b/uvdat/core/models/file_item.py
@@ -24,7 +24,7 @@ class FileItem(TimeStampedModel):
         pass
 
 
-@receiver(models.signals.pre_delete, sender=FileItem)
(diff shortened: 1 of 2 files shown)


uvdat/core/models/file_item.py
```diff
@@ -21,3 +22,9 @@ class FileItem(TimeStampedModel):
     def download(self):
         # TODO: download
         pass
+
+
+@receiver(models.signals.pre_delete, sender=FileItem)
```
jjnesbitt, line 27: Wouldn't we want these signals to be `post_delete` instead of `pre_delete`? For example, if there was an exception during the delete (say some check/unique constraint violation), the file would be gone in minio but the row would still exist.

annehaley: Good point. Changed in [4ba2bcb](https://github.com/OpenGeoscience/uvdat/pull/83/commits/4ba2bcb2555639adfc25b537c293cd181698fbdf).

jjnesbitt (approved):

PR #84, opened 2024-10-17
Update Github Actions

annehaley:


.github/workflows/ci.yml
```diff
@@ -6,44 +6,60 @@ on:
...
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
         with:
-          compose-file: |
-            docker-compose.yml
-            docker-compose.override.yml
-      - run: |
-          docker compose exec django tox -e check-migrations
+          python-version: "3.11"
```
jjnesbitt, line 46: Since our `setup.py` sets the python version as `>=3.10`, I think we should use `3.10` here, to ensure we're testing things accurately. Bumping that version can be done separately.

.github/workflows/ci.yml
```diff
@@ -6,44 +6,60 @@ on:
...
+      - name: Install tox
+        run: |
+          pip install --upgrade pip
+          pip install tox
+      - name: Install GDAL
+        run: |
+          sudo apt-add-repository ppa:ubuntugis/ubuntugis-unstable
```
jjnesbitt, line 53: I don't love that we're using the `unstable` PPA for GDAL here. I'm guessing this is due to the [stable ppa](https://launchpad.net/~ubuntugis/+archive/ubuntu/ppa) not containing an image for Ubuntu 24.04 (noble). I think the way to solve this is to pin the ubuntu image we run on to `ubuntu-22.04`, rather than try to switch to `24.04` and use unstable repositories.

annehaley: Good idea. Updated in [08271ba](https://github.com/OpenGeoscience/uvdat/pull/84/commits/08271ba9d3e1471cdeafbf8b4888ccebde774df4)

.github/workflows/ci.yml
annehaley: Ok, changed in [08271ba](https://github.com/OpenGeoscience/uvdat/pull/84/commits/08271ba9d3e1471cdeafbf8b4888ccebde774df4)

jjnesbitt (approved):
Looks great!

PR #85, opened 2024-10-25
Flood Timeseries Dataset

### LINKED ISSUE #75: Revisit time-varying raster dataset (flood simulation)
Use `large-image` to composite all time steps into a single file, access frames separately.

annehaley:
Resolves #75.

This PR adds the simulated flood timeseries dataset created by Jack Watson. This involves the addition of a custom ingest function to split a single zip file item into many tiff file items. The current implementation is not abstracted; it relies on the presence of `vrt` files in the zip and relies on the file names to assert ordering.

Additionally, this PR adds animation controls to the current layer slider so that the user can automatically step through the layers on a dataset. The video (2x speed) below demonstrates this behavior on the flood timeseries dataset.

https://github.com/user-attachments/assets/68d8352d-f502-4407-8c22-263eae605f21

### DIFF
--- a/uvdat/core/tasks/map_layers.py
+++ b/uvdat/core/tasks/map_layers.py
@@ -49,7 +49,8 @@ def add_styling(geojson_data, style_options):
 
 def split_raster_zip(file_item, style_options):
     """For each raster image in a zip file, create a new FileItem."""
-    # NOTE: This implementation is not abstract; this was built for the Boston Flood Timeseries dataset.
+    # NOTE: This implementation is not abstract;
+    # this was built for the Boston Flood Timeseries dataset.
     import large_image_converter
 
     if file_item.file_type == 'zip':
--- a/uvdat/core/tasks/map_layers.py
+++ b/uvdat/core/tasks/map_layers.py
@@ -66,7 +66,7 @@ def split_raster_zip(file_item, style_options):
                 archive_file.write(file_item.file.open('rb').read())
                 # unzip all
                 with zipfile.ZipFile(archive_path) as zip_archive:
-                   zip_archive.extractall(extracted_path)
+                    zip_archive.extractall(extracted_path)
             # read VRT files
             vrt_files = list(extracted_path.glob('**/*.vrt'))
             vrt_files.sort()  # rely on filenames for ordering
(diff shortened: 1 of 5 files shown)


uvdat/core/tasks/map_layers.py
```diff
@@ -46,8 +47,61 @@ def add_styling(geojson_data, style_options):
     return geopandas.GeoDataFrame.from_features(features)
 
 
+def split_raster_zip(file_item, style_options):
+    """For each raster image in a zip file, create a new FileItem."""
+    # NOTE: This implementation is not abstract;
+    # this was built for the Boston Flood Timeseries dataset.
+    import large_image_converter
+
+    if file_item.file_type == 'zip':
```
jjnesbitt, line 56:
```suggestion
    if file_item.file_type != 'zip':
        return
```

uvdat/core/tasks/map_layers.py
```diff
@@ -46,8 +47,61 @@ def add_styling(geojson_data, style_options):
     return geopandas.GeoDataFrame.from_features(features)
 
 
+def split_raster_zip(file_item, style_options):
+    """For each raster image in a zip file, create a new FileItem."""
+    # NOTE: This implementation is not abstract;
+    # this was built for the Boston Flood Timeseries dataset.
```
jjnesbitt, line 53: If this is the case, I think the function should be labeled as such.

uvdat/core/tasks/map_layers.py
```diff
@@ -46,8 +47,61 @@ def add_styling(geojson_data, style_options):
...
+                    new_file_item.file.save(cog_filepath, ContentFile(cog.read()))
+                if cog_size > 0:
+                    create_raster_map_layer(new_file_item, style_options)
+
+
 def create_raster_map_layer(file_item, style_options):
     """Save a RasterMapLayer from a FileItem's contents."""
+    if style_options is None:
+        style_options = {}
+    if file_item.file_type == 'zip':
```
jjnesbitt, line 101: Similar to my other comment, if the logic for extracting this data is specific to the boston flood timeseries dataset, a comment here explaining that would be good. That way, in the future, we know this is not general behavior.

jjnesbitt (changes requested):

annehaley:
After a team discussion about changing the relationships between Projects, Datasets, and Layers, we decided that this implementation will soon be non-applicable. This PR will be closed and revisited after the design change.

PR #87, opened 2024-11-06
Improve Testing Infrastructure

jjnesbitt:
This PR improves the testing infrastructure of the UVDAT API, through the addition of pytest fixtures, factory-boy factories, and some core tests. These tests aren't fully expansive, and don't cover every possible scenario, but instead attempt to test the core business logic that is already implemented. Many of the default views that we leverage from DRF are _not_ explicitly tested, as we do very little (if any) modification of those viewset behaviors, and as such we would simply be testing DRF itself.

Some side effects of this PR:

1. Fixed a bug where the dataset list would only show datasets that were used in projects a user was added to.
2. Fixed a bug that allowed collaborators to change the owner of a project.
3. The removal of many unused endpoints.


uvdat/core/tests/conftest.py
```diff
@@ -1,12 +1,29 @@
 from django.contrib.auth.models import User
 from django.contrib.gis.geos import Point
 import pytest
-from pytest_factoryboy import register
 from rest_framework.test import APIClient
 
 from uvdat.core.models import Project
 
-from .factories import UserFactory
+from .factory_fixtures import *  # noqa: F403, F401
```
annehaley, line 8: Is it more appropriate to be explicit here about which classes are imported? Or is there a particular reason you used the star import?

uvdat/core/tests/test_dataset.py
```diff
@@ -0,0 +1,168 @@
...
+    factory = (
+        vector_map_layer_factory
+        if dataset_type == Dataset.DatasetType.RASTER
+        else raster_map_layer_factory
+    )
+    for _ in range(3):
+        factory(dataset=dataset)
+
+    # Check that nothing is returned, since map_layers will only
+    # return map layers that match the dataset's type
```
annehaley, line 114: I'm not sure that this is intended behavior, but this is fine for now. We will change this soon to allow multimodal datasets.

uvdat/core/tests/conftest.py
jjnesbitt: The import exists solely to allow for pytest to pick up the defined factory fixtures when running tests. We would do nothing with these if imported.

uvdat/core/rest/project.py
```diff
@@ -48,6 +58,7 @@ def simulation_results(self, request, **kwargs):
         simulation_results = project.simulation_results.all()
         return HttpResponse(simulation_results, status=200)
 
+    # TODO: This should be a POST
```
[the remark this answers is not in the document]
jjnesbitt: Yes, that was added as a note to address outside of this PR.

PR #90, opened 2024-11-13
Add precommit

### LINKED ISSUE #88: Use precommit
To streamline development and avoid mistakes, we should use precommit to enforce the following:

- [ ] Commit messages should include semantic versioning tags
- [ ] Linting and typing tests should pass

annehaley:
Resolves #88.

This PR adds `.pre-commit-config.yaml` to the top-level of this repository so that we can run pre-commit checks during development. These checks include linting with `flake8` and confirming that the commit message follows [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) structure (so that our semantic release CI job can interpret them).

To install:
```
pip install pre-commit
pre-commit install --hook-type commit-msg
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-added-large-files
+  - repo: https://github.com/pycqa/flake8
+    rev: 7.1.0
+    hooks:
+      - id: flake8
+        additional_dependencies:
+          [
+            flake8-black,
+            flake8-bugbear,
+            flake8-docstrings,
+            flake8-isort,
+            flake8-quotes,
+            pep8-naming,
+          ]
+        args: ["--ignore=E123,E203,W503,D10"]
+        exclude: "sample_data"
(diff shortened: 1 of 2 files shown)


BryonLewis (approved):
This functions properly and I don't see any technical issues.

Not a blocker and it could be that I'm just not as familiar with convential-pre-commit and it's use.

Just curious as to the reason behind the convential-pre-commit.  It makes sense in a highly structured environment where you are specifically taking out and doing extremely small bit-size tasks and issues.  In most of our projects we are sometimes implementing an entire feature in a week or two.  During that process on the branch we're working on there may be multiple commits.  On my other projects I squash merge and use PR's as history because they typically contain contextual information (PR Description + comments/discussions).  I didn't look into it but is there an easy way to escape out of the convential-pre-commit rules?  If not I know that personally I would just be making less commits when working on a PR (instead of at least every day when I'm done working).

Just to be clear, I know this the correct way to be doing things on a software project I'm just wondering if it will eat into feature velocity.

annehaley:
> Just curious as to the reason behind the convential-pre-commit

@BryonLewis  We use the `python-semantic-release` Github Action when we merge to master. This action depends on interpreting commit messages to determine an appropriate release bump (major, minor, patch). We've been trying to remember to include the conventional type tags in our commit messages (feat, fix, refactor, test, etc.) and now `convential-pre-commit` will help us stay consistent.

PR #91, opened 2024-12-02
Integrate `tile2net` task

annehaley:
This PR adds an orthoimagery (3.6 GB raster) dataset to the Boston floods use case and adds a new simulation type which runs `tile2net` on a selected imagery dataset. The "Segment Curbs" task downloads tiles of raster data in a format expected by `tile2net` and then runs the `tile2net` `generate` and `inference` functions. The output files generated by the inference step are then interpreted as new Datasets to add to the current project. 

The newly-created vector datasets are sent through our existing vector conversion process and are made available to view on the map. The first Dataset is the "polygons" generated by the inference, which represent the segmentation of curbs, crosswalks, and other walkable paths in the imagery. The second Dataset is the "network" generated by the inference, which is a processed version of the "polygons" set, consisting of semi-connected line segments. An avenue for future work would be to interpret the "network" set into our own `Network` representation. However, due to the nature of walkable urban systems, the resulting `Network` would be very disconnected and not viable for meaningful GCC analysis.

Results
- Clifton Park set got poor results, took 2 mins for a 144 MB file  (not trained for this area, not very walkable) 
image

- Boston set got good results, took 2 hours 10 mins for 3.6 GB file (trained for this area, semi-walkable)
image

- Capitol Hill set got fair results, took 8 mins for a 1 GB file (trained for this area, highly walkable park)
image


uvdat/core/tasks/map_layers.py
```diff
@@ -59,64 +60,83 @@ def create_raster_map_layer(file_item, style_options):
...
-            output_data.write(band, 1)
-            output_data.close()
-
-            cog_raster_path = Path(temp_dir, 'cog_raster.tiff')
-            large_image_converter.convert(str(raster_path), str(cog_raster_path))
-            with open(cog_raster_path, 'rb') as cog_raster_file:
-                new_map_layer.cloud_optimized_geotiff.save(
-                    cog_raster_path, ContentFile(cog_raster_file.read())
+        try:
+            source = large_image.open(raw_data_path)
```
jjnesbitt, line 71: I think this function could use some comments as well. It's not really clear to someone reading the code, what the intent of these code blocks are.

sample_data/use_cases/boston_floods/composite_orthoimagery.py
jjnesbitt, line 1: Where is this file invoked? I don't see it used anywhere.

jjnesbitt (changes requested):

web/src/layers.ts
```diff
@@ -505,14 +505,13 @@ export function styleRasterDatasetLayer(
...
-    defaultStyle?.palette ||
-    "terrain") as string;
+    defaultStyle?.palette) as string;
   const colormapRange = options?.colormap?.range || defaultStyle?.data_range;
   const nodataValue = options.nodata || defaultStyle?.transparency_threshold;
 
   const tileParams: Record<string, string> = {
     projection: "EPSG:3857",
-    band: "1",
+    // band: "1",
```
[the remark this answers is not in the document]
annehaley: Thanks for pointing this out. I had commented this out for the (now default) case of using no colormap for a raster, and I didn't test out the colormap functionality. We do need this argument when applying a colormap, so I added this back in for the colormap case in [5511b25](https://github.com/OpenGeoscience/uvdat/pull/91/commits/5511b25db56e52d3cf6e1e632273e03f1b752209).

uvdat/core/tasks/map_layers.py
```diff
@@ -48,6 +48,7 @@ def add_styling(geojson_data, style_options):
 
 def create_raster_map_layer(file_item, style_options):
```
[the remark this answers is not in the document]
annehaley: Gotcha, I split this function in [f493c8d](https://github.com/OpenGeoscience/uvdat/pull/91/commits/f493c8da3bda444d3d5c579bdfd5da8a0c946a91)

uvdat/core/tasks/map_layers.py
annehaley: Good idea, I added comments in [f493c8d](https://github.com/OpenGeoscience/uvdat/pull/91/commits/f493c8da3bda444d3d5c579bdfd5da8a0c946a91)

sample_data/use_cases/boston_floods/composite_orthoimagery.py
annehaley: You're right, I don't invoke this code anywhere. This is a script that I invoked manually to composite separate raster files into one before using the output file for ingest. I want to keep this code somewhere, though, in case we want to incorporate it into the ingest process more formally later. But that will be dependent on our modelling decisions relating to composite rasters.

jjnesbitt: Maybe this could either be made into a management command, or documented in some way, so that if time passes the inputs and outputs of this script can still be known.

jjnesbitt (changes requested):
Just left one more comment about the `composite_orthoimagery` script. Otherwise this LGTM

sample_data/use_cases/boston_floods/composite_orthoimagery.py
annehaley: Now that I've had some time to explore the new model design in #94, I was able to confirm that we can perform spatial compositiing dynamically (from multiple separate files). Now I know that we won't need this script for our conversion process, as I had been concerned about before. I removed the file in [e3f1f48](https://github.com/OpenGeoscience/uvdat/pull/91/commits/e3f1f48cc51fa05870350e9d5eefd1e9c2803faa), but we can always come back to this PR for the code if it is needed later. 

(If that day ever comes, the usage is very simple. Just put all the target rasters in a folder and change the value for `folder` to point to that path. Then run `python composite_orthoimagery.py` and the result file, `geospatial_composite.tiff`, will be in the target folder.)

PR #94, opened 2025-01-08
Models Redesign

annehaley:
This PR adds the `Layer` and `LayerFrame` models to support new Dataset structures such as timeseries and multilayer Datasets. These changes are limited to server implementation, so the client application is incompatible on this branch. We have an upcoming client redesign effort that will complement these changes on a new branch. In the time between, we can use a simplified interface to review data with this new structure (#95).

The models implemented in this PR follow this design diagram:
image


uvdat/core/tests/test_dataset.py
```diff
@@ -90,58 +90,49 @@ def test_rest_dataset_gcc(
...
-    # return map layers that match the dataset's type
-    resp = authenticated_api_client.get(f'/api/v1/datasets/{dataset.id}/map_layers/')
+    resp = authenticated_api_client.get(f'/api/v1/datasets/{dataset.id}/layers/')
     assert resp.status_code == 200
-    assert resp.json() == []
+
+    data: list[dict] = resp.json()
+    assert len(data) == 3
+
+    print(data)
```
jjnesbitt, line 108:
```suggestion
```

uvdat/core/tasks/regions.py
```diff
@@ -51,6 +55,6 @@ def create_source_regions(vector_map_layer, region_options):
     # Save updated features to layer
     new_geodata = geopandas.GeoDataFrame.from_features(new_feature_set).set_crs(3857)
     new_geodata.to_crs(4326)
-    vector_map_layer.write_geojson_data(new_geodata.to_json())
-    vector_map_layer.save()
-    print('\t', f'{region_count} regions created.')
+    vector_data.write_geojson_data(new_geodata.to_json())
+    vector_data.save()
```
jjnesbitt, line 59: This feels like a vestige of the old design. Why would creating source regions pull data from this field, and the overwrite it afterwards? Isn't the point of a separate `Region` model to avoid this? This also makes the function not idempotent.

jjnesbitt (changes requested):

sample_data/use_cases/new_york_energy/import_networks.py
```diff
@@ -24,9 +24,15 @@ def get_metadata(feature):
 
 def create_network(dataset, network_name, geodata):
     print(f'\t\tCreating network for {network_name}.')
+    vector_data = VectorData.objects.create(
+        name=network_name,
+        dataset=dataset,
+    )
```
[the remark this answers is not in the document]
annehaley: This `create_network` function is a special implementation for the NY Energy use case. When we ingest this network, it may come directly from NYSDP API calls (if not cached), and therefore the vector data does not have any source file.

uvdat/core/models/data.py
```diff
@@ -41,8 +34,12 @@ def get_image_data(self, resolution: float = 1.0):
             return data.tolist()
 
 
-class VectorMapLayer(AbstractMapLayer):
-    geojson_file = S3FileField(null=True)
+class VectorData(models.Model):
+    name = models.CharField(max_length=255, default='Vector Data')
+    dataset = models.ForeignKey(Dataset, related_name='vectors', on_delete=models.CASCADE)
+    source_file = models.ForeignKey(FileItem, null=True, on_delete=models.CASCADE)
+    geojson_data = S3FileField(null=True)
```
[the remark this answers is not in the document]
annehaley: Sometimes we can create data objects programmatically (like from API calls, or as part of a task like `tile2net`) instead of from source files. The source file field is optional if that relationship exists (hence `null=True` on that field) but the relationship to a Dataset is mandatory.

uvdat/core/models/networks.py
```diff
@@ -43,7 +44,11 @@
 
 
 class Network(models.Model):
+    name = models.CharField(max_length=255, default='Network')
     dataset = models.ForeignKey(Dataset, on_delete=models.CASCADE, related_name='networks')
```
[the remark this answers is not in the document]
annehaley: I originally thought that I might answer in a similar manner to the question about data objects, but I think at the Network level, it does make sense to make the `vector_data` field mandatory and remove the extra `dataset` field. I think it's reasonable to require that networks have a relationship to a VectorData object. Changed in [11f7cde](https://github.com/OpenGeoscience/uvdat/pull/94/commits/11f7cde767d1b0564acf24e343b8850726ec4bff).

uvdat/core/models/regions.py
```diff
@@ -1,11 +1,15 @@
 from django.contrib.gis.db import models as geo_models
 from django.db import models
 
+from .data import VectorFeature
 from .dataset import Dataset
 
 
 class SourceRegion(models.Model):
```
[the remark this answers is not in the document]
annehaley: I agree. I had been thinking about the model as just "Region" for a while now but hesitated to make that change without your input. Changed in [750c135](https://github.com/OpenGeoscience/uvdat/pull/94/commits/750c135480b6b9d595047c21e23df82255f87968).

uvdat/core/rest/data.py
```diff
@@ -60,20 +56,21 @@
         core_vectorfeature t,
         bounds
     WHERE
-        t.map_layer_id = %(map_layer_id)s
+        t.vector_data_id = %(vector_data_id)s
         AND ST_Intersects(
             ST_Transform(t.geometry, %(srid)s),
             ST_Transform(bounds.geom, %(srid)s)
         )
+        FILTERS
```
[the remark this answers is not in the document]
annehaley: String replacement was the best solution I could come up with for adding N extra filters to the SQL query. The vector tile request accepts a set of filters (key-value pairs) from the query parameters and replaces this "FILTERS" marker with the string of N extra lines with AND statements. For every key-value pair in the request parameters, we add the following lines to the SQL string:
```
AND t.properties::jsonb ? '{key}'
AND t.properties::jsonb ->> '{key}' = '{value}'
```
These lines check for the existence of the key in the vector feature properties and check that the value in the properties matches the value in the request filter. This functionality allows us to filter by properties for different frames referencing the same VectorData. We can also use this filtering functionality for advanced styling later.

Since you were the one to write this SQL string, I am open to changing how this is done with your guidance.

uvdat/core/tests/test_populate.py
```diff
@@ -42,11 +44,14 @@ def test_populate():
     assert Project.objects.all().count() == 2
     assert Dataset.objects.all().count() == 4
     assert FileItem.objects.all().count() == 7
+    # TODO use correct expected counts
+    assert Layer.objects.all().count() == -1
+    assert LayerFrame.objects.all().count() == -1
```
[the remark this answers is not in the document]
annehaley: Thanks for catching this. The reason is simply that I forgot to come back to this TODO. I was holding off on running the longer tests and made this note to myself so I could run the populate test later and change the expected counts. Changed in 03ba49b.

uvdat/core/tests/test_dataset.py
annehaley: Changed in [3c24d8c](https://github.com/OpenGeoscience/uvdat/pull/94/commits/3c24d8c48fc58be5a0410c29469a55a1d5f9c59c).

uvdat/core/tasks/regions.py
annehaley: We do the same thing when we create a network from vector data. The purpose of overwriting the vector data is just to update the properties of vector features. In the lines above this highlighted section, for each vector feature representing a region, we add a few properties: `region_id`, `region_name`, and `dataset_id`. We use these vector feature properties client-side to make requests that reference the Region objects. We do the same for networks since we need node and edge IDs.

uvdat/core/rest/data.py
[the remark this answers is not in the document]
jjnesbitt: For what it's worth:

1. You don't need the first `t.properties::jsonb ? '{key}'` clause in that filter. If the json field doesn't have `key` as a key, it will simply fail the filter, and not be included in the result set.
2. `properties` is already a `jsonb` field, so the `::jsonb`cast isn't necessary.
3. I'm not sure if this applies, but if we intent do to nested filters (e.g. `a.b.c == 'sometext'`), this filter won't allow that. You need to use the `#>>` operator for that, and I think it would be relatively easy to generalize that behavior to both flat and nested keys. 

Aside from that, I think this approach makes sense. I would only advocate that the text `FILTERS` be replaced with some sentinel value, like `REPLACEME_WITH_FILTERS` or something. Just something that would be obvious to someone when debugging. `FILTERS` on its own could be mistaken as a reserved postgres keyword, if you weren't familiar with it.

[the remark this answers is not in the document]
annehaley: Good point, I hadn't considered nested properties. I updated the filtering logic in [81dffa1](https://github.com/OpenGeoscience/uvdat/pull/94/commits/81dffa181cfb94b9453c82423e89d7f87c15cd4e) to use a recursive function for the creation of the filter string. I tested this out with a set of simple filters like `{frame: 0, attrs: {status: 'good'}}`. I incorporated your suggestions for the filter clause into this function.

uvdat/core/rest/data.py
```diff
@@ -60,20 +57,42 @@
...
+            try:
+                value = ast.literal_eval(value)
+            except ValueError:
+                pass
+            if isinstance(value, dict):
+                string = get_filter_string(value, string, key_prefix + key)
+            else:
+                key_path = '{' + key_prefix + key + '}'
+                string += f"\n\t\tAND t.properties #>> '{key_path}' = '{value}'"
+    return string
```
jjnesbitt, line 90: I think this is too complicated and presents a maintainability concern. I think it's fair to define the following constraints:

1. Nested keys must be specified with javascript syntax (e.g `a.b.c`).
2. Filter key/value pairs must be a primitive type. Objects are not supported.
3. To distinguish between strings and other primitive types, string values must be quoted in strings when passed to the backend (e.g. `"1"` vs `1`).

Given those constraints, what do you think about this simplified version:

```suggestion
def get_filter_string(filters: dict | None = None):
    if filters is None:
        return ''

    return_str = ''
    for key, value in filters.items():
        key_path = key.replace('.', ',')
        return_str += f" AND t.properties #>> '{{{key_path}}}' = {value}"
    return return_str

```

To be honest even this is slight overkill for what we are planning to support at the moment, but it will at least allow basic filtering and nested property filtering.

annehaley: I think we have to include the single quotes around the value. When I use this function and test the vector tile endpoint with a query of `frame=0`, the following error is raised:
```
django.db.utils.ProgrammingError: operator does not exist: text = integer                                   
LINE 50:          AND t.properties #>> '{frame}'  = 0                                                                                                               
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

jjnesbitt: Ah true, the `#>>` operator itself returns a `text` field. I guess for now we'll have to just use strings. In that case, I've updated the above suggestion. It does make the implementation easier, at least.

annehaley: Ok great, I'll commit this suggestion.

annehaley: I've updated the `jupyter-explorer` branch to be compatible with these changes: [52352cf](https://github.com/OpenGeoscience/uvdat/pull/96/commits/52352cf9c71a559d05d8d8e86830514a8247bd84).

jjnesbitt (approved):

PR #100, opened 2025-02-10
Fix edge cases with create_{vector,raster}_map_layer

floryst:
- `style_options` can be None, since we don't have schema validation for input `datasets.json`
- Passing in a vector filename with an unsupported extension results in `geojson_data` not being defined.


jjnesbitt (approved):
LGTM

Issue #69, opened 2024-09-25
Improve codebase in the wake of MapLibre-GL migration


jjnesbitt:
There are several things that remain less than ideal immediately following the merging of https://github.com/OpenGeoscience/uvdat/pull/63, which should be addressed.

1. The complexity and multi-purpose role of [`updateVisibleMapLayers`](https://github.com/OpenGeoscience/uvdat/blob/efc9a4fae3113ebad2bbcc712794e63595a01fff/web/src/layers.ts#L702)
2. The handling of datasets with [multiple networks](https://github.com/OpenGeoscience/uvdat/blob/efc9a4fae3113ebad2bbcc712794e63595a01fff/web/src/utils.ts#L186)
3. The handling of network datasets when [populating simulation results](https://github.com/OpenGeoscience/uvdat/blob/efc9a4fae3113ebad2bbcc712794e63595a01fff/web/src/components/SimulationsPanel.vue#L104)
4. The handling of datasets in [`toggleNodeHandler'](https://github.com/OpenGeoscience/uvdat/blob/efc9a4fae3113ebad2bbcc712794e63595a01fff/web/src/components/map/MapTooltip.vue#L137)
5. The default value for [circle layer opacity](https://github.com/OpenGeoscience/uvdat/blob/efc9a4fae3113ebad2bbcc712794e63595a01fff/web/src/layers.ts#L104)

There are further things that could be addressed at a more architectural level, but these were the immediate things to fall out of the maplibre implementation.

PR #102, opened 2025-02-11
UI updates

annehaley:
This PR implements the new user interface design by @faiza-a ([Figma link](https://www.figma.com/design/OQ7RqVWPSlsyNMcLeI4j69/UVDAT?node-id=5-4&p=f&t=8W6Qcuhkp8qkoO6j-0)). These design updates are intended to be compatible with the model updates in #94, allowing a user to interact with Layers on a Dataset and Frames on a Layer. This new design is also intended to bring a new look and feel to the UVDAT webapp.

In the course of implementing this UI redesign, major refactoring was necessary to adhere to the new database models design. The logic functions for map layers, styling, and networks are now simplified and isolated into their own modules. Several unnecessary store variables were removed. 

Watch a demo video of this new design [here](https://drive.google.com/file/d/16y2QV7VBJebnvN5xy05b3CJx99YIjcfa/view?usp=sharing).


annehaley:
@jjnesbitt As you review, I'd like to get your input as to whether this PR resolves #69.

web/src/components/DatasetsPanel.vue
```diff
@@ -0,0 +1,206 @@
...
+          <v-expansion-panel-text>
+            <v-treeview
+              :items="groupDatasets"
+              item-value="id"
+              item-title="name"
+              item-children="layers"
+              open-on-click
+            >
+              <template v-slot:title="{title, item}">
+                <div class="dataset-title">
```
jjnesbitt, line 119: Having the whole row clickable seems more intuitive to me. Since there's no other action that can be taken on that row, forcing the user to click a small icon seems unnecessary.

```suggestion
                <div class="dataset-title" @click="toggleSelected([item])">
```

web/src/components/DatasetsPanel.vue
```diff
@@ -0,0 +1,206 @@
...
+                    ></v-icon>
+                    <v-checkbox-btn
+                      v-else-if="item.layers && props.selectedIds"
+                      :model-value="props.selectedIds.includes(item.id)"
+                      @click.stop="() => toggleSelected([item])"
+                      style="display: inline"
+                    />
+                    {{ title }}
+                  </div>
+                  <div v-if="item.layers">
```
jjnesbitt, line 137: This should  be `item.layers.length`, but if I do that, I get console errors about accessing `.length` on undefined. Why would `Dataset.layers` ever be undefined?

jjnesbitt (changes requested):

web/src/components/DatasetsPanel.vue
jjnesbitt: Also it seems the class `dataset-title` is misleading, since this is actually displaying a layer, not a dataset.

web/src/components/SideBars.vue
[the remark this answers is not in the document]
annehaley: Unfortunately, yes. The new Figma design does not include that list. We may need to add a dedicated "Network" panel to bring that back in the future, along with some other Network-specific things. I'll note this for future iterations of Faiza's design.

web/src/components/DatasetsPanel.vue
```diff
@@ -0,0 +1,206 @@
...
+                      color="primary"
+                      v-tooltip="'Add to Selected Layers'"
+                      @click="() => toggleSelected([item])"
+                    ></v-icon>
+                    <v-checkbox-btn
+                      v-else-if="item.layers && props.selectedIds"
+                      :model-value="props.selectedIds.includes(item.id)"
+                      @click.stop="() => toggleSelected([item])"
+                      style="display: inline"
+                    />
```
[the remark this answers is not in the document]
annehaley: There are two reasons that this strangeness is necessary. 

The first is that this highlighted section is within a `v-treeview` component, where `item` can be either a Dataset or a Layer. The top-level items in the treeview are Datasets, and each Dataset item has one or more children, which are the Dataset's Layers. So the statement that evaluates `item.layers` is actually a check to determine whether `item` is a Dataset (because a Layer would not have the `layers` property). Thus, if `!item.layers`, we know the item is a Layer and we show the plus icon next to it. 

The second is that this component is used in more than one place. We use this component in the main Datasets panel within the sidebar, but we also use it twice within the Project config page for adding/removing Datasets in a Project. When we use this component in the Project config page, we pass an extra prop called `selectedIds` to keep track of which Datasets the user has selected for adding/removing. So we check for the existence of `props.selectedIds` to determine whether we should show the checkboxes. Only items that are Datasets get the checkbox in this mode, and the items that are Layers do not show a plus icon in this mode.

web/src/components/DatasetsPanel.vue
annehaley: See above comment; `item` may not be a Dataset within this treeview.

web/src/components/DatasetsPanel.vue
annehaley: Ok, I changed the structure (and class name) of the item title here to make the whole row clickable for adding a layer to selected layers: [52dc78d](https://github.com/OpenGeoscience/uvdat/pull/102/commits/52dc78da5aaf709d058d5442a02b0a707682db32). It required a bit of restructuring because we don't want the whole row to be clickable in the case where a Dataset has a checkbox next to it in the Project config context. In that case, clicking on the row expands the Dataset to show its Layers, while clicking on the checkbox toggles the selected state.

web/src/components/DatasetsPanel.vue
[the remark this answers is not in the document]
jjnesbitt: If using the same component for these two different parts of the UI is truly necessary, I would suggest that we at least have some comments in the template that indicate where that occurs. 

However, I'll also just say that I don't think having some duplicated code in two different spots is problem, as while there may be some functionality overlap, the ability to for the component to operate in the one specific way it's intended makes it much more readable and extensible. For example, I don't really see a reason to have the dataset selection in the project config panel as a treeview. it's probably too much information. I think the dataset granularity is sufficiently "fine", and would make the code simpler for that use case, while also relieving the other `DatasetsPanel.vue` component of the burden of the project config use case.

This could be made more ergonomic by having folders for views/components that use several vue files, like what the `components/map` folder does. In this case, you would probably have something like:

```
components/
├─ ProjectConfig/
│  ├─ ProjectConfig.vue
│  ├─ AccessControl.vue
│  ├─ DatasetsPanel.vue
```

This could be done for other files/folders too, if it applies. 

Anyway, if none of this sticks, I'd still at least like to see the combined UI component commented to indicate the divergent behavior.

web/src/components/DatasetsPanel.vue
jjnesbitt: This is some more anecdotal evidence towards what I'm proposing in  https://github.com/OpenGeoscience/uvdat/pull/102#discussion_r1978298094, FWIW.

web/src/components/FloatingPanel.vue
```diff
@@ -0,0 +1,209 @@
...
+            v-tooltip="'Close Panel'"
+            @mousedown="closePanel"
+          ></v-icon>
+        </div>
+        <v-card-text class="pa-2">{{ panel.label }}</v-card-text>
+        <v-card-text v-if="!panel.collapsed" class="pa-2 panel-content">
+          <DatasetsPanel v-if="props.id === 'datasets'" :datasets="availableDatasets"/>
+          <LayersPanel v-else-if="props.id === 'layers'"/>
+          <ChartsPanel v-else-if="props.id === 'charts'"/>
+          <AnalyticsPanel v-else-if="props.id === 'analytics'"/>
```
jjnesbitt, line 125: Is there a strong reason not to simply make this a vue slot, where you'd provide the component itself from `Sidebars.vue`? That way, `FloatingPanel.vue` wouldn't have to know exactly what it's wrapping, and might be more extensible.

jjnesbitt: I'm confused, since the two seem functionally equivalent? These component are only used in this one spot in the template. If it simply can't be done, that's fine, maybe I'm just missing something.

annehaley: That commit just changed how the floating panels are used within each sidebar. Instead of a predefined order within each sidebar (where each floating panel in the DOM knows what its contents should be and uses a slot), we now use a `v-for` in each sidebar, where the number of floating panels and the content of each one is dynamic. We could still use a slot, but then we'd just have this branching `v-if` logic in two places (within the `v-for`s in each sidebar) instead of one.

jjnesbitt: I see your point now, feel free to revert that change.

faiza-a (commented):
I reviewed UI implementation based on [Finalized UI Mockups](https://www.figma.com/design/OQ7RqVWPSlsyNMcLeI4j69/UVDAT?node-id=5-4&p=f&t=otlwGy0ciHKqEXYp-0). The review points are in this [Figma File](https://www.figma.com/design/OQ7RqVWPSlsyNMcLeI4j69/UVDAT?node-id=949-3667&p=f&t=otlwGy0ciHKqEXYp-0). The review addressed some of the alignment, colors, spacing etc improvements. All pointers i added in Figma file so far is addressed by Anne in this PR and seems to be implemented as expected.
I will be creating more UI improvement pointers in Figma for next PR.

Thanks @annehaley for addressing the UI refinements. Nice work! 🚀

web/src/layerStyles.ts
```diff
@@ -0,0 +1,177 @@
...
+            let source = map.getSource(mapLayer.source) as RasterTileSource;
+            if (source?.tiles?.length) {
+                const oldQuery = new URLSearchParams(source.tiles[0].split('?')[1])
+                const newQuery = new URLSearchParams(getRasterTilesQuery(style));
+                if(newQuery.toString() !== oldQuery.toString()) {
+                    source.setTiles(source.tiles.map((url) => url.split('?')[0] + '?' + newQuery))
+                }
+            }
+        }
+    }
```
jjnesbitt, line 84: There is nothing we can do here if `mapLayer` is undefined:
```suggestion
    const mapLayer = map.getLayer(mapLayerId);
    if (mapLayer === undefined) {
        return;
    }

    if (mapLayerId.includes("fill")) {
        map.setPaintProperty(mapLayerId, 'fill-opacity', opacity / 2);
        map.setPaintProperty(mapLayerId, 'fill-color', color);
    } else if (mapLayerId.includes("line")) {
        map.setPaintProperty(mapLayerId, 'line-opacity', opacity);
        map.setPaintProperty(mapLayerId, 'line-color', color);
    } else if (mapLayerId.includes("circle")) {
        map.setPaintProperty(mapLayerId, 'circle-opacity', opacity);
        map.setPaintProperty(mapLayerId, 'circle-stroke-opacity', opacity);
        map.setPaintProperty(mapLayerId, 'circle-color', color);
        map.setPaintProperty(mapLayerId, 'circle-stroke-color', color);
    } else if (mapLayerId.includes("raster")) {
        map.setPaintProperty(mapLayerId, "raster-opacity", opacity)
        let source = map.getSource(mapLayer.source) as RasterTileSource;
        if (source?.tiles?.length) {
            const oldQuery = new URLSearchParams(source.tiles[0].split('?')[1])
            const newQuery = new URLSearchParams(getRasterTilesQuery(style));
            if(newQuery.toString() !== oldQuery.toString()) {
                source.setTiles(source.tiles.map((url) => url.split('?')[0] + '?' + newQuery))
            }
        }
    }
```

jjnesbitt (changes requested):
Just one style suggestion but otherwise looks good to me!

jjnesbitt (approved):

PR #103, opened 2025-02-14
Additional API updates

annehaley:
In the course of continuing UI updates on #102, I found some more server-side changes that are needed to use the new Network model in our API. This includes an update to the new `TokenAuth` class and a new set of viewpoints for Networks. The client will now leverage these endpoints for Network-related logic (rather than going through Dataset endpoints, since the relationship between a Network and a Dataset is no longer one-to-one).

### DIFF
--- a/uvdat/core/admin.py
+++ b/uvdat/core/admin.py
@@ -88,10 +88,7 @@ class RegionAdmin(admin.ModelAdmin):
 
 
 class NetworkAdmin(admin.ModelAdmin):
-    list_display = ['id', 'category', 'get_dataset_name']
-
-    def get_dataset_name(self, obj):
-        return obj.dataset.name
+    list_display = ['id', 'category']
 
 
 class NetworkEdgeAdmin(admin.ModelAdmin):
--- a/uvdat/core/rest/networks.py
+++ b/uvdat/core/rest/networks.py
@@ -1,0 +1,34 @@
+from drf_yasg.utils import swagger_auto_schema
+from rest_framework import serializers
+from rest_framework.decorators import action
+from rest_framework.response import Response
+from rest_framework.viewsets import ModelViewSet
+
+from uvdat.core.models import Network
+from uvdat.core.rest.access_control import GuardianFilter, GuardianPermission
+from uvdat.core.rest.serializers import NetworkSerializer
+
+
+class GCCQueryParamSerializer(serializers.Serializer):
+    exclude_nodes = serializers.RegexField(r'^\d+(,\s?\d+)*$')
+
+
+class NetworkViewSet(ModelViewSet):
+    queryset = Network.objects.all()
+    serializer_class = NetworkSerializer
+    permission_classes = [GuardianPermission]
+    filter_backends = [GuardianFilter]
+    lookup_field = 'id'
+
+    @swagger_auto_schema(query_serializer=GCCQueryParamSerializer)
+    @action(detail=True, methods=['get'])
+    def gcc(self, request, **kwargs):
+        network = self.get_object()
+
+        # Validate and de-serialize query params
+        serializer = GCCQueryParamSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        exclude_nodes = [int(n) for n in serializer.validated_data['exclude_nodes'].split(',')]
+
+        gcc = network.get_gcc(excluded_nodes=exclude_nodes)
+        return Response(gcc, status=200)
--- a/uvdat/core/rest/networks.py
+++ b/uvdat/core/rest/networks.py
@@ -13,6 +13,10 @@ class GCCQueryParamSerializer(serializers.Serializer):
     exclude_nodes = serializers.RegexField(r'^\d+(,\s?\d+)*$')
 
 
+class GCCResultSerializer(serializers.Serializer):
+    gcc = serializers.ListField(child=serializers.IntegerField())
+
+
 class NetworkViewSet(ModelViewSet):
     queryset = Network.objects.all()
     serializer_class = NetworkSerializer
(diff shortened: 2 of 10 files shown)


uvdat/core/tasks/networks.py
```diff
@@ -25,6 +25,7 @@ def create_network(vector_data, network_options):
     dataset = vector_data.dataset
     Network.objects.filter(vector_data__dataset=dataset).delete()
     network = Network.objects.create(
+        name=dataset.name + ' Network',
```
jjnesbitt, line 28: Are we currently constraining Networks to be one-to-one with datasets?

uvdat/core/rest/networks.py
```diff
@@ -0,0 +1,34 @@
...
+    def gcc(self, request, **kwargs):
+        network = self.get_object()
+
+        # Validate and de-serialize query params
+        serializer = GCCQueryParamSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        exclude_nodes = [int(n) for n in serializer.validated_data['exclude_nodes'].split(',')]
+
+        gcc = network.get_gcc(excluded_nodes=exclude_nodes)
+        return Response(gcc, status=200)
```
jjnesbitt, line 34: It's probably best to use a serializer here, but I understand this is just relocated code. Not a huge deal imo.

uvdat/core/tasks/networks.py
annehaley: No, do you think that we should add numbering to this name string to account for multi-network datasets?

annehaley: Added numbering to avoid name conflicts in [93f7e45](https://github.com/OpenGeoscience/uvdat/pull/103/commits/93f7e45e1f5705654db1336af76d4308209cb908)

uvdat/core/rest/networks.py
annehaley: Created a custom serializer to validate gcc results in [98e1d2b](https://github.com/OpenGeoscience/uvdat/pull/103/commits/98e1d2b720467f3cbfcf359cdd202f1141320f5b)

uvdat/core/admin.py
```diff
@@ -88,10 +88,7 @@ def get_dataset_name(self, obj):
 
 
 class NetworkAdmin(admin.ModelAdmin):
-    list_display = ['id', 'category', 'get_dataset_name']
-
-    def get_dataset_name(self, obj):
-        return obj.dataset.name
+    list_display = ['id', 'category']
```
zachmullen, line 91: Why remove the dataset name? Is it not something administrators would want to see in this view?

uvdat/core/rest/tokenauth.py
```diff
@@ -7,6 +7,9 @@
 
 class TokenAuth(OAuth2Authentication):
     def authenticate(self, request):
+        auth = super().authenticate(request)
+        if auth is not None:
+            return auth
         token = request.query_params.get('token')
         if token is None:
             token_string = request.headers.get('Authorization')
```
zachmullen, line 15: I know this logic didn't come from this PR, but it seems like this will break anytime a request is made that doesn't contain an `Authorization` header.

uvdat/core/admin.py
annehaley: The previous model updates PR removed the relationship to the Dataset (instead inferring that relationship through the associated VectorData object), and this serializer method should have been removed then. Removing it now because I got errors when trying to access the Networks page. 

Also, when we create Network objects in our ingestion method, the name comes from the Dataset: `name=f'{dataset.name} Network {existing.count() + 1}',`.

uvdat/core/rest/tokenauth.py
annehaley: When I try to send a basic request with no headers, I just get a 401 response:
```
>>> r = requests.get('http://localhost:8000/api/v1/datasets/')
>>> r.status_code
401
```
Still, I know that this class could probably be hardened. Lmk if we can make any changes to improve its logic.

zachmullen: Ah ok, I guess it doesn't even reach this code unless the header has been passed. That's fine then.

uvdat/core/rest/networks.py
```diff
@@ -0,0 +1,40 @@
...
+
+        # Validate and de-serialize query params
+        serializer = GCCQueryParamSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        exclude_nodes = [int(n) for n in serializer.validated_data['exclude_nodes'].split(',')]
+
+        gcc = network.get_gcc(excluded_nodes=exclude_nodes)
+        result = GCCResultSerializer(data=dict(gcc=gcc))
+        if result.is_valid():
+            return Response(result.data.get('gcc'), status=200)
```
jjnesbitt, line 40: While we're unlikely to hit this case, since this is a trivial serializer, but if the result _weren't_ valid, this endpoint would error, as it's returning nothing.

I think it's better to explicitly invoke this case, like so:

```suggestion
        result.is_valid(raise_exception=True)
        return Response(result.validated_data['gcc'], status=200)
```

This will raise an exception if the data input is malformed.

jjnesbitt: I'll also point out that generally speaking the serializer defines the response shape, so you can imagine having a response structure of something like:

```
{
  "count": 10,
  "gcc": [
    1,
    2,
    ...
  ]
} 
```

that you would just pass to the response like so:

```
return Response(result.validated_data)
```

This is probably beyond the scope of this PR, but I just thought I'd mention it.

annehaley: Ok, I applied your suggested change. And I see what you mean about the return value, but for this PR I tried not to change the response structure so it would not affect the client. We can return to that later if necessary.

jjnesbitt (approved):

PR #117, opened 2025-03-13
Project Permissions: Remove users

### LINKED ISSUE #109: Project permissions: Remove user not working
The implementation of [`Project.set_permissions`](https://github.com/OpenGeoscience/uvdat/blob/438fbecd87305c9a00d22e7f3942044a23c12e01/uvdat/core/models/project.py#L75) does not account for removing users.

annehaley:
Resolves #109

### DIFF
--- a/uvdat/core/models/project.py
+++ b/uvdat/core/models/project.py
@@ -4,7 +4,7 @@ from django.contrib.auth.models import User
 from django.contrib.gis.db import models as geo_models
 from django.db import models, transaction
 from guardian.models import UserObjectPermission
-from guardian.shortcuts import assign_perm, get_users_with_perms
+from guardian.shortcuts import assign_perm, get_users_with_perms, remove_perm
 
 from .dataset import Dataset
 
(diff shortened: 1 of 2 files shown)


uvdat/core/models/project.py
```diff
@@ -60,13 +60,17 @@ def set_owner(self, user: User):
         assign_perm('owner', user, self)
 
     @transaction.atomic()
-    def add_collaborators(self, users: list[User]):
+    def set_collaborators(self, users: list[User]):
+        for user in get_users_with_perms(self, only_with_perms_in=['collaborator']):
+            remove_perm('collaborator', user, self)
```
floryst, line 65: I just want to check my understanding here: this for loop removes collaborator permissions from users, but then in the next line we delete all user permissions associated with this object. How are the two different?

annehaley: This first loop iterates through users that already have collaborator permissions on the target project and removes the collaborator perm. The second loop iterates through users in the new list and removes any other permissions they might have on the target project (for example, if a user in the new collaborator list was previously a follower, the follower perm should be removed before the collaborator perm is added). The difference between these two loops is the user list they iterate on: old collaborators list versus new collaborators list

annehaley: We previously thought that `self.delete_users_perms(users)` would account for removing permissions, but we failed to consider that in the case of removing user permissions, the `users` list would simply lack the users that should be removed. So the addition of the loop above is intended to consider the differences between the old user list and the new one.

willdunklin (approved):

floryst (approved):

PR #118, opened 2025-03-19
Analytics

### LINKED ISSUE #28: Refactor Simulation Function Listing / Handling
I think this is getting to the point where we should probably consider refactoring how we handle storing available simulations, so that we don't complicate things further. I don't want to block this PR on it, but I think after this feature is in, we could do some refactor behind the scenes, preserving feature parity. What do you think?

_Originally posted by @AlmightyYakob in https://github.com/OpenGeoscience/uvdat/pull/23#discussion_r1342946834_

### LINKED ISSUE #67: Improve "Available Simulations" endpoint
For some reason, the endpoint used to retrieve available simulations per context is taking exceptionally long (around 20 seconds). This is at the endpoint:

```
http://localhost:8000/api/v1/simulations/available/context/{context_id}
```

It seems to make 75 SQL queries, which indicates to me that something unintentional is going on. This should be investigated and addressed.

### LINKED ISSUE #108: Simulations -> Analytics
In order to upgrade UVDAT's current simulation system, we should address the following items:

- [ ] Rename SimulationType to AnalysisType and SimulationResult to AnalysisResult
- [ ] Refactor definition and fetching of AnalysisTypes (#67, #28)
- [ ] Use Django channels for push notifications, show a progress bar and text updates to the user
- [ ] Report total runtime when a task finishes
- [ ] Add "Flood Simulation" task that selects a data product from Northeastern. User inputs include "precipitation hyetograph" (Chart object depicting parabolic or NRCS Type 2), and "likelihood" (4% chance / 1 in 25 year or 1% chance / 1 in 100 year), and "time period" (2030-2050 or 2081-2100). We can expand the options later, but for now we'll proceed with two choices for each of the three inputs, resulting in 8 permutations. Northeastern agreed to provide data products for these 8 permutations, where each data product is a zip of tiff files. Each tiff file represents one hour in a 24 hour range, starting with the beginning of the precipitation event. The output of this task is the data product that corresponds to the selected inputs. Eventually this will be reimplemented to run their code and generate new data as output. If possible, we should report "certainty" as an additional output.
- [ ] Improve "Flood Scenario 1" task. Maybe rename to "Network Flood Analysis" or "Flood Network Outages"?  Instead of the "flood" input being any VectorData object with category "flood", the user must select a result from a "Flood Simulation" run, which would have a Northeastern data product as output. Add a user inputs called "depth tolerance" and "station radius (m)". Use depth values in data products, and mark a node as deactivated if depth is over tolerance within the station radius. Instead of ordering outages by elevation, order first by timesteps in the data product, then by elevation.
- [ ] Rename "Recovery Scenario" task to "Network Recovery". Add a "none" recovery strategy to reflect natural recovery. For outputs, in addition to the node recovery animation, this task should also output a static GCC chart that is not affected by user interactions in the client. Add an additional visualization mode where users can color network nodes by importance. Node importance is determined by the change in GCC when that node becomes deactivated.

annehaley:
Resolves #108. Resolves #67. Resolves #28.

This PR accomplishes the following:
- Create AnalysisResult model and remove SimulationResult model
- Remove all server-side references to "Simulations"
- Define AnalysisTypes with class-based implementation, all inheriting from an abstract base class
- Refactor rest viewset for analytics
- Implement "Flood Simulation" task that outputs pregenerated data products from Northeastern
- Rename and reimplement "Flood Network Failure" task now using Northeastern flood data products and improved logic
- Refactor "Network Recovery" and "Segment Walkable Paths" tasks
- Update Analytics panel in web client
- Remove all client-side references to "Simulations"
- Update Network animations to change with the flood dataset frames
- Improve performance and style of network animations
- Use Django channels for push notifications
- Add GCC chart output to Network Recovery task

[This video](https://drive.google.com/file/d/16cl9MTuOP3A-A1EZngnGygk2Wu0RRIZt/view?usp=sharing) demonstrates the above functionality.


uvdat/asgi.py
```diff
@@ -1,11 +1,31 @@
...
 
-application = get_asgi_application()
+application = ProtocolTypeRouter(
+    {
+        'http': get_asgi_application(),
+        'websocket': AuthMiddlewareStack(
+            URLRouter(
+                [
+                    path(
+                        'api/v1/analytics/project/<int:project_id>/results/',
```
zachmullen, line 23: You may want to consider not putting this URL under your REST API namespace since it's a websocket connection designed to service a specific web client.

annehaley: Good point, changed in [18f4ff8](https://github.com/OpenGeoscience/uvdat/pull/118/commits/18f4ff86795a2dc7bd72f010ea2f9a629a289aaf)

docker-compose.yml
```diff
@@ -42,7 +42,15 @@ services:
     logging:
       driver: none
 
+  redis:
+    image: redis:latest
+    volumes:
+      - redis-data:/redis/data
+    ports:
+      - 6379:6379
```
floryst, line 50: Is it necessary to port forward this to localhost if you're REDIS_URL is already using the container hostname?

uvdat/core/tasks/analytics/analysis_type.py
```diff
@@ -0,0 +1,19 @@
+from abc import ABC, abstractmethod
+
+
+class AnalysisType(ABC):
+    def __init__(self, *args):
+        self.name = None
+        self.description = None
+        self.db_value = None  # cannot be longer than 25 characters
```
floryst, line 8: It might be good to put this as a member field with a comment so that it can show up in IDEs. Runtime enforcement might be better, unless silent truncation is inconsequential.

docker-compose.yml
annehaley: You're right, it's not necessary. Removed in [c223aed](https://github.com/OpenGeoscience/uvdat/pull/118/commits/c223aed192059457a79266ba771507ffd5843463)

uvdat/core/tasks/analytics/analysis_type.py
annehaley: Django handles runtime enforcement because any attempt to save an AnalysisResult object with a value for `analysis_type` over 25 characters will raise a `ValidationError`.

dev/.env.docker-compose-native
```diff
@@ -7,3 +7,4 @@ DJANGO_MINIO_STORAGE_ACCESS_KEY=minioAccessKey
 DJANGO_MINIO_STORAGE_SECRET_KEY=minioSecretKey
 DJANGO_STORAGE_BUCKET_NAME=django-storage
 DJANGO_HOMEPAGE_REDIRECT_URL=http://localhost:8080/
+REDIS_URL=redis://redis:6379
```
jjnesbitt, line 10: Since this is the `-native` file, these should be available on `localhost`.
```suggestion
REDIS_URL=redis://localhost:6379
```

You'll also want to add back in the port forwarding on the redis service in `docker-compose.yml`.

jjnesbitt (changes requested):

dev/.env.docker-compose-native
annehaley: Changed in [12b65cb](https://github.com/OpenGeoscience/uvdat/pull/118/commits/12b65cbb9e66200b9bb72f6d0b117118bc640826)

tox.ini
```diff
@@ -46,11 +46,13 @@ passenv =
     DJANGO_MINIO_STORAGE_ENDPOINT
     DJANGO_MINIO_STORAGE_SECRET_KEY
     DJANGO_HOMEPAGE_REDIRECT_URL
+    REDIS_URL
 extras =
     dev
     test
 commands =
     pip install large-image-converter --find-links https://girder.github.io/large_image_wheels
+    pip install large-image[gdal] --find-links https://girder.github.io/large_image_wheels
```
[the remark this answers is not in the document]
annehaley: You're right; the tests that need these libraries are actually already marked as slow and are run in the nightly job. I just needed to remove these install commands in [32b19a9](https://github.com/OpenGeoscience/uvdat/pull/118/commits/32b19a9f0053bfae943a2adb4a649a13bf28dc99)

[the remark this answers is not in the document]
jjnesbitt: I think those two lines should be moved into the run command of a new tox section (`test-slow`) that would run the slow tests. That way, whether I'm running tox natively, or through docker-compose, or in CI, I can just run `tox -e test-slow` and I'll get the same result (as long as GDAL is installed). Currently, the native workflow is left out of this, as it's installed in both the dev dockerfile and nightly CI separately.

web/src/layers.ts
```diff
@@ -42,11 +44,43 @@ export function getDBObjectsForSourceID(sourceId: string) {
                     else if (frame.raster) DBObjects.raster = frame.raster;
                 }
             })
+            DBObjects.network = availableNetworks.value.find((n) => n.vector_data == DBObjects.vector?.id)
         }
     })
     return DBObjects
 }
 
+export function addLayer(layer: Layer) {
```
jjnesbitt, line 53:
```suggestion
// Add this layer to selectedLayers, which will then trigger updateLayersShown to add it to the map
export function addLayer(layer: Layer) {
```

jjnesbitt (approved):
Aside from my recent comment suggestion, it seems you've resolved my concerns, so seems fine by me.

tox.ini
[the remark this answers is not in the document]
annehaley: Changed in [8bb6031](https://github.com/OpenGeoscience/uvdat/pull/118/commits/8bb60317bd8d2cae41dd1dce898182a58c8f63d5)

floryst (approved):

Issue #110, opened 2025-03-12
Metadata View


annehaley:
Most UVDAT Models (Dataset, FileItem, Chart, RasterData, VectorData, Layer, LayerFrame, Network, NetworkNode, NetworkEdge, and Region) have a JSON field "metadata" to contain any pertinent information that des not belong in other fields. The web client does not yet have a way to show this information to the user. We need a dialog that will display this information using the `RecursiveTable` component we already have in our client, as well as a way to open this dialog for any objects of the types listed above.

annehaley:
`FileItem`, `RasterData`, `VectorData`, and `Network` have no current visual representation in the interface, and `NetworkNode`, `NetworkEdge`, and `Region` display metadata in the map tooltip. Thus, the scope of this issue will be limited to creating a metadata display for `Dataset`, `Chart`, `Layer`, and `LayerFrame`.

annehaley:
@willdunklin I created a branch `metadata-examples` for developing the metadata viewer. Just checkout the branch and run `python manage.py metadata_test` to create objects with example metadata, then you can switch back to your branch and continue development.

PR #120, opened 2025-03-24
Add MetadataView modal

willdunklin:
Addresses #110.

Adds MetadataView modal for `Dataset`, `Layer`, `Chart`, and `LayerFrame`.

Example for metadata view buttons (vertical elipses). These buttons appear for all metadata-viewable elements:
dataset selection metadata buttons

Example of LayerFrame metadata:
layerframe metadata

Example of larger metadata modal, the modal displays nested `RecursiveTable` components, which can display arbitrary JSON objects:
dataset metadata


annehaley (changes requested):
A great start, but I have the following UI suggestions:
- the modal dialog should have an X close button in the top right corner rather than a "Close" button at the bottom (so we don't have to scroll to the end of the table). You can copy the X button from the Projects Config modal.
- the title of the dialog should be "Metadata For [name]" rather than just the object name
- the width of the metadata table should be limited to the width of the dialog so there is no horizontal scrollbar.
- Can you interpret and format date strings?
- The Chart metadata view button should be on the list item, not the chart display (I shouldn't need to select the chart to view its metadata)

web/src/components/MetadataView.vue
```diff
@@ -0,0 +1,43 @@
+<script setup lang="ts">
+  import { ref, computed } from 'vue';
+  import RecursiveTable from './RecursiveTable.vue';
+
+  const props = defineProps<{
+    metadata: Object | undefined; // TODO: make less generic (Object)
```
annehaley, line 6: You can use `Record<string, any>` as a more specific typing

annehaley (changes requested):
Thank you for addressing my comments. The date parsing looks good. I just have a few more trivial suggestions below.

annehaley (approved):
Looks good, thank you for addressing my comments.

PR #123, opened 2025-03-25
Show tooltips below layer name

floryst:
The "Add to Selected Layers" button tooltip covered the layer name. Move the tooltip to the bottom as to not obscure the layer name.

image

This PR has no dependency on #122, but it is stacked on top so that I didn't have to stash my other changes.

### DIFF
--- a/web/src/components/sidebars/DatasetsPanel.vue
+++ b/web/src/components/sidebars/DatasetsPanel.vue
@@ -63,15 +63,20 @@ function toggleSelected(items: Layer[]) {
               class="item-title"
             >
               <div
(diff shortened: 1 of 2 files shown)


willdunklin (approved):

annehaley (commented):
Since you had mentioned that the light mode tooltips should have a higher contrast background, you could change the value for `surface-variant` and `on-surface-variant` for the light theme in `themes.ts`. I have no preference on those colors, so you can experiment with what looks best to you.

annehaley (approved):

PR #133, opened 2025-05-05
Gracefully handle missing `Authorization` header

jjnesbitt:
Currently if you make a request to an endpoint that uses the `TokenAuth` class from swagger (or just your browser), it throws a `500 Internal Server Error`, because it expects the `Authorization` header to be present unconditionally.


jjnesbitt:
I've realized this only addresses part of a larger problem, so I will close this in favor of that PR (not yet opened).

PR #135, opened 2025-05-05
Limit scope of ipyleaflet token authentication class

jjnesbitt:
Depends on #134

The `TokenAuth` class was doing two things that somewhat complicate the indented usage and scope of an authentication class.

1. It was inheriting from and calling first the `OAuth2Authentication` class, which means that the first thing this `TokenAuth` class did during `authenticate` was attempt to do oauth authentication. This is peculiar as it means that this class is actually performing two completely separate forms of authentication, depending on the conditions. Since `OAuth2Authentication` is already specified in `settings.py`, we should simply rely on that instead.
2. It attempted to perform token authentication if the `token` query param was missing. This is also unexpected, and means we're again performing two different forms of authentication. Since we specify the `TokenAuthentication` class in `settings.py`, we should rely on that class in cases of proper token auth.

This PR fixes the above issues by removing abstraction from `OAuth2Authentication`, as well as the implementation of standard token authentication. The scope is now limited to just the `token` query parameter, returning `None` otherwise (allowing any other authentication classes to attempt auth, if necessary).


uvdat/core/rest/explorer.py
```diff
@@ -0,0 +1,22 @@
...
+
+        try:
+            user = User.objects.get(auth_token=token)
+        except User.DoesNotExist:
+            return None
+
+        if not user.is_active:
+            return None
+
+        return (user, None)
```
jjnesbitt, line 22: @annehaley The docstring for the base `authenticate` function states the following:

```
Authenticate the request and return a two-tuple of (user, token)
```

Is there a particular reason we return `None` here instead of the token?

annehaley: There isn't a functional reason for returning `None` instead of the token. I had been following [this SO answer](https://stackoverflow.com/a/32846841) for a custom auth method. In their answer, there is no token so they return `(user, None)`. Let's update this to return the token.

annehaley (approved):

PR #136, opened 2025-05-06
DetailView: Download related files

### LINKED ISSUE #114: FileItem View / Download
We should add a dialog to list the FileItem objects associated with a Dataset. Each FileItem in the list should display its name, description, upload datetime, uploading user, download button, and metadata.  We may also want to expand a FileItem to see the VectorData and/or RasterData objects that were created from it, as well as similar details and download options for those objects.

annehaley:
Resolves #114


jjnesbitt:
It seems there's no base condition when clicking on related files / objects, leading to a situation where I can attempt to open the object I started out with, and cause an error. We should track which objects are "parents" to the object we're currently viewing, and exclude them from the related objects list.

https://github.com/user-attachments/assets/6c265b28-c24b-4ff7-bb3a-e6b45aca8809

web/src/components/DetailView.vue
```diff
@@ -0,0 +1,176 @@
...
+      <div v-if="related" style="min-width: 500px;">
+        <v-card-subtitle>Related {{ relatedLabel }} </v-card-subtitle>
+        <v-list :items="related" item-value="id" >
+          <template v-slot:title="{ item }">
+            <div v-if="item" v-tooltip="item.name">{{ item.name }}</div>
+          </template>
+          <template v-slot:subtitle="{ item }">
+          </template>
+          <template v-slot:append="{ item }">
+            <DetailView v-if="item" :details="item"/>
```
jjnesbitt, line 151: I see what you mean about having dialogs on top of each other.

I wonder if using this component recursively isn't the answer here, but rather, having a "stack" of parent objects that you would push to. The "current" object would be the one at the top of the stack, clicking on the 3 dot details would "push" that new object to the top of the stack, and we'd have a back navigation arrow to "pop" the current object off the stack and look at the previous one (we could use sideways transitions to make the pushing and popping obvious to the user). 

That way, everything is contained in one dialog, it's more intuitive to the user, and you can easily reference parent objects (kind of ties into my other comment about checking parent objects to prevent a loop).

Does what I'm proposing make sense? If so what do you think?

annehaley: Great idea, I implemented the stack in [f77e91b](https://github.com/OpenGeoscience/uvdat/pull/136/commits/f77e91bfcb8c07c5aaee60c493d6e4cf2a1b1961) and the slide transition in [b390b27](https://github.com/OpenGeoscience/uvdat/pull/136/commits/b390b2724272133fdb2cd95e50b08f6863b7cec3). Lmk if this is what you were imagining.

jjnesbitt: I've got a few critiques still, but it's definitely much improved.

1. It might be useful to show other details of a dataset besides metadata, like ID, description, and category. Maybe this could also apply to other objects, I'm not sure.
2. We might want to change the width of the dialog from `min-content` to some fixed value or something. Having all of that data in a tall thin strip on the screen is not very valuable.
3. We might also want to set a max-height, for similar reasons.
4. It seems the first "jump" doesn't actually perform a transition. I thought this might have been related to treating the first value differently from the rest of the stack, which is why I made the change in [f843e34](https://github.com/OpenGeoscience/uvdat/pull/136/commits/f843e340a4f2ba867a0600137caf60a14fe6b6fc), but it seems that didn't fix it. However, I still think that change is useful. If you disagree feel free to revert it.
5. Closing the dialog should reset the state. Currently if I re-open the same dataset, I'm still nested where I previously was, which is confusing.
6. To me it would make more sense to show the raster/vector data first, and _then_ show related files, since I would think raster/vector data is more relevant to the user than the underlying files. However, that would require a decent re-work of this PR, and I don't know ultimately how much that really matters.

https://github.com/user-attachments/assets/69aa5cd2-f9ff-4b1e-a538-94fc41b5d141




7. It seems something is going on with the recursive table component and displaying metadata in certain circumstances. I've included a video below that shows how it's treating strings as arrays in the commuter rail vector metadata.

https://github.com/user-attachments/assets/b8921d25-3291-4cfc-aa39-f02aef79923b

jjnesbitt: Great work!

jjnesbitt (approved):

PR #138, opened 2025-05-08
Use Sets to represent network nodes and edges

jjnesbitt:
Since we perform a lot of operations with network nodes and edges, storing them as an array can be a bit cumbersome both mentally and from a performance standpoint (since simply finding a node in a list requires an O(n) operation). Sets allow for quicker lookups, but more importantly, they also allow for more intuitive and performant operations between groups of nodes. For example, if you wanted to find the common nodes between two groups of nodes, with arrays, it would be something like:

```javascript
const commonNodes = groupA.filter((node) => groupB.includes(node));
```

which has quadratic time complexity. With a set, the code is:

```javascript
const commonNodes = groupA.intersection(groupB);
```

which is of linear time complexity (and easier to read imo).

This extends to more complex examples that are easy to articulate with set functions, but become complicated when working with two arrays.

### DIFF
--- a/web/src/api/rest.ts
+++ b/web/src/api/rest.ts
@@ -110,12 +110,14 @@ export async function getNetworkEdges(networkId: number): Promise<NetworkEdge[]>
 export async function getNetworkGCC(
   networkId: number,
   exclude_nodes: number[]
-): Promise<number[]> {
-  return (
+): Promise<Set<number>> {
+  const nodes = (
     await apiClient.get(
       `networks/${networkId}/gcc?exclude_nodes=${exclude_nodes.toString()}`
     )
   ).data;
+
+  return new Set(nodes);
 }
 
 export async function getRasterDataValues(rasterId: number): Promise<RasterDataValues> {
--- a/web/src/api/rest.ts
+++ b/web/src/api/rest.ts
@@ -87,16 +87,28 @@ export async function getDatasetLayers(datasetId: number): Promise<Layer[]> {
   return (await apiClient.get(`datasets/${datasetId}/layers`)).data;
 }
 
+
+function processNetwork(data: Record<string, any>): Network {
+  return {
+    ...data,
+    nodes: new Set(data.nodes),
+    edges: new Set(data.edges),
+  } as Network
+}
+
 export async function getDatasetNetworks(datasetId: number): Promise<Network[]> {
-  return (await apiClient.get(`datasets/${datasetId}/networks`)).data;
+  const { data } = (await apiClient.get(`datasets/${datasetId}/networks`)).data;
+  return data.map((network: any) => processNetwork(network));
 }
 
 export async function getProjectNetworks(projectId: number): Promise<Network[]> {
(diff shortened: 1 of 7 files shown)


web/src/layerStyles.ts
```diff
@@ -139,17 +139,19 @@ export function styleNetwork(network: Network) {
...
-                    const inactive = network.deactivated?.nodes.filter((n) => (
-                        !deactivate?.includes(n) && !activate?.includes(n)
-                    )) || [];
-                    let gcc = network.gcc || []
+
+                    const deactivate = Array.from(network.changes?.deactivate_nodes || []);
+                    const activate = Array.from(network.changes?.activate_nodes || []);
+                    const inactive = Array.from(network.deactivated?.nodes || []).filter((n) => (
+                        !network.changes?.deactivate_nodes?.has(n) &&
+                        !network.changes?.activate_nodes?.has(n)
```
jjnesbitt, line 147: @annehaley I'm particularly interested to know if this change affects behavior at all, as I'm not sure if I'm able to test this functionality locally atm.

web/src/api/rest.ts
```diff
@@ -87,16 +87,28 @@ export async function getDatasetLayers(datasetId: number): Promise<Layer[]> {
...
 
+
+function processNetwork(data: Record<string, any>): Network {
+  return {
+    ...data,
+    nodes: new Set(data.nodes),
+    edges: new Set(data.edges),
+  } as Network
+}
+
```
annehaley, line 98: Since this function doesn't make a rest call, I think we should move it to networks.ts

web/src/networks.ts
```diff
@@ -22,44 +22,47 @@ export async function toggleNodeActive(
...
-        edges: []
+        nodes: new Set(),
+        edges: new Set()
     }
     if (animation) {
         network.changes = {
-            deactivate_nodes: nodeIds.filter((n) => !network.deactivated?.nodes.includes(n)),
-            activate_nodes: network.deactivated.nodes.filter((n) => !nodeIds.includes(n))
+            deactivate_nodes: nodeIds.difference(network.deactivated.nodes),  // nodes about to be deactivated
+            activate_nodes: network.deactivated.nodes.difference(nodeIds),    // nodes about to be reactivated
```
annehaley, line 47: This line raises `TypeError: CallSetMethodIfWrapped method called on incompatible Proxy` because `network.deactivated.nodes` is a proxy for a set.

annehaley (changes requested):

web/src/layerStyles.ts
annehaley: I also cannot test this due to the error mentioned below; once that is resolved I can evaluate this.

annehaley: Testing the non-animated version with the networks panel, however, revealed to me that I should have moved the selected clause to the top (so it would be evaluated first and override any styles related to activation state). I'll add that commit to my branch

jjnesbitt:
It seems this in infeasible due to how `Set` is implemented (and it's incompatibility with `Proxy`).

PR #152, opened 2025-06-10
Pre-load all frames in multi-frame layer

jjnesbitt:


jjnesbitt:
It seems we will not address this for the moment, due to implementation challenges.

PR #154, opened 2025-06-12
Install all large-image sources in Dockerfile

annehaley:
Instead of only installing the GDAL, PIL, and Mapnik sources in our Dockerfile (used for our Django and Celery containers), this PR changes the Dockerfile to install all large-image sources.


jjnesbitt (approved):
I don't really notice the slight improvement w.r.t raster tiling you mentioned after rebuilding my containers with this, but it works properly nonetheless.

PR #155, opened 2025-06-12
Add a "test" use case

annehaley:
This PR creates a new use case called "test". Running `python manage.py populate test` will create a new Project called "Test" with two datasets. The two datasets have the same generated vector data, but one is multiframe and the other is a single frame.


jjnesbitt (approved):
Very concise and functional, nice!

PR #169, opened 2025-07-21
Reuse existing layer copies if not in `selectedLayers`

### LINKED ISSUE #156: Adding layer again after removing from selected layers causes copy to be created
To reproduce, perform the following steps:

1. Add a layer to the selected layers panel
2. Remove that layer from the selected layers panel by clicking the X
3. Add the same layer again to the selected layers panel

What you'll notice is that on step 3, even though that layer wasn't in selected layers when it was added, it was added as a copy (with a `(1)` suffix). This seems to be a bug, or if it is intentional behavior, is it rather confusing.

annehaley:
Resolves #156

### DIFF
--- a/web/src/store/layer.ts
+++ b/web/src/store/layer.ts
@@ -130,7 +130,19 @@ export const useLayerStore = defineStore('layer', () => {
     });
     if (existing.length) {
       copy_id = existing.length;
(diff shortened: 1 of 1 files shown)


annehaley:
Closing in favor of #173

PR #170, opened 2025-07-21
Update populate script

### LINKED ISSUE #107: Prevent broken downloads during dataset ingest
There's an bug that I've encountered when running `populate`. This pertains to the new [redesign](https://github.com/OpenGeoscience/uvdat/tree/redesign-staging).

1. Try to run populate (I encountered the bug with the boston floods dataset, but it could happen with any of them)
2. The download ends abruptly, writing an incomplete zip file into minio, but nonetheless continues
3. The `Dataset` and associated `FileItem` objects are created (the file item pointing to the corrupt zip file in minio)
4. When `convert_file_item` is run, reading the zip file fails, raising an exception and ending the `populate` script
5. When trying again, it skips creating the dataset again or running `ingest_file`, because the dataset already exists (occurs [here](https://github.com/OpenGeoscience/uvdat/blob/f6d42f79b89327d72eb29a690ccbadca0f3dbcab/sample_data/ingest_use_case.py#L128-L145))
6. Any further attempts fail in the same way, because it keeps fetching the corrupt zip file in minio and proceeding from there

There are at least two things that can be done to improve the resiliency of the `populate` script. 

1. Wrap [`ingest_datasets`](https://github.com/OpenGeoscience/uvdat/blob/f6d42f79b89327d72eb29a690ccbadca0f3dbcab/sample_data/ingest_use_case.py#L119) (or maybe a new singular `ingest_dataset` function that handles a specific dataset only) in a transaction. This would ensure that if an exception is raised in the process of ingesting the dataset, the results aren't saved, and don't continue to pollute the app.
2. Add hashes for each of the files downloaded from `datasets.json`. This would ensure that if a download only partially completes, we could either retry a fixed amount of times, or raise an exception that the file contents are mismatched (or both).

### LINKED ISSUE #157: Populate script replacement/redownload
It would be nice to have the populate management script to be able to replace data locally by replacing the data in the ./sampledata/downloads location or you can trigger it to redownload the data from the path URL.

This is mainly useful for testing data imports frequently when testing performance or other geojson imports.

annehaley:
Using [pooch](https://github.com/fatiando/pooch) for downloading data files uses file hashes to ensure that downloaded files are correct and complete. Resolves #107.

Adding the `--no_cache` argument to the populate command allows the user to force deletion of existing Datasets, FileItems, and Charts and create new ones. Resolves #157.

### DIFF
--- a/sample_data/ingest_use_case.py
+++ b/sample_data/ingest_use_case.py
@@ -2,6 +2,7 @@ from datetime import datetime
 import importlib
 import json
 import os
+import pooch
 from pathlib import Path
 
 from django.contrib.auth.models import User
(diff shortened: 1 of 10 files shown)


jjnesbitt (approved):
Seems to work as intended

BryonLewis (approved):
I'm not going to pull and check.  I believe Jake has already done that and the intended functionality works.  I 

Below are questions about support for different use cases that could be a separate PR if it doesn't fit in this PR.

I'm curious if this means it's a requirement to have a URL link?
Could you add a new dataset and not require it to be hosted elsewhere online?  I.E place it in the ./download directory with the proper path+filename?

Along those lines is there an ability to replace a fileItem and it's layers completely without having to redownload?  I'm only asking because it DIVERS-H when using a new data format like netCDF or fmv support it's fairly common to import multiple times to see how the data ingestion works.  I wouldn't want to redownload a 1.3GB image/video file from data.kitware.com  each time.

annehaley:
@BryonLewis I think that [c6e51fe](https://github.com/OpenGeoscience/uvdat/pull/170/commits/c6e51fe28e2e11add5d143115c5261a62ac415ef) can cover the use case you're describing. You can now specify a path for an existing local file without a url, and pooch will not be used in that case. The local file would be used as-is.

As for your second question, (assuming a url is specified) the download won't occur in the `--no-cache` case if the hash of the local copy matches the `known_hash`. Using `--no-cache` will still delete the FileItem and its Layers from the DB, but the download only occurs if the local copy doesn't exist or doesnt match the `known_hash`

PR #174, opened 2025-07-25
Standardize handling of source and layer IDs

jjnesbitt:
Our current method of parsing layer and source IDs is done ad hoc everywhere it's needed. This PR introduces well defined functions for parsing source and layer IDs into their constituent parts.

### DIFF
--- a/web/src/store/layer.ts
+++ b/web/src/store/layer.ts
@@ -21,6 +21,30 @@ function uniqueLayerIdFromLayer(layer: Layer) {
   return `${layer.id}.${layer.copy_id}`;
 }
 
+/**
+ * Note: Rasters also have an extra `bounds` source, which allows for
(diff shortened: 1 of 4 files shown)


annehaley (commented):
Functionally this looks good, but I have an organizational suggestion. The functions related to these strings (`uniqueLayerIdFromLayer`, `sourceIdFromLayerFrame`, `parseLayerString`, `parseSourceString`, `sourceIdFromMapLayerId`) should probably all be defined in the same place. Their implementations are linked, so if we ever change one, we would need to change them all, and doing so would be easier if they were all together. It doesn't matter to me whether you put them in the layer store or the map store, but I think that they should be together given their semantic proximity.

annehaley (approved):
Thanks for reorganizing. LGTM!

PR #176, opened 2025-07-28
Layer Styles: Allow different style for each layer copy

### LINKED ISSUE #171: Don't switch back to default style when advanced styling menu is closed
A user should be able to compare two different styles on two copies of the same layer. We should allow style selection to persist when the advanced styling menu is closed. We can still revert unsaved changes on the selected style, though.

annehaley:
Resolves #171.
Rather than switching back to the default style every time the styling menu is closed, this PR changes that behavior, allowing the current selected style to persist even if it isn't the default for that layer. This allows users to apply different styles to different copies of the same layer for comparison.

Some notes:
- We still discard any unsaved changes when the styling menu is closed
- Added a tooltip to the gear icon which displays the name of the current style applied to that layer copy. This will help the user keep their layer copies straight without needing to open the styling menu. If no style is applied, the tooltip reads "Configure styling".
- In order to persist this information after the styling menu is closed, some refactoring was necessary. Previously, `styleStore.selectedLayerStyles` contained a mapping of keys to `StyleSpec` objects. Now, we store the whole `LayerStyle` object, which contains a `StyleSpec`.

The following video demonstrates this feature:
https://github.com/user-attachments/assets/ff497d4a-575a-46d0-8de6-f35d50678949


jjnesbitt (changes requested):
One comment but otherwise looks good

jjnesbitt (approved):

PR #177, opened 2025-07-28
Fix inconsistent source/layer ID generation

jjnesbitt:
Follow up fix to #174 

1. Rasters were adding layers as `tile` and `mask` instead of `raster` and `fill` (the maplibre layer type)
1.`bounds` was being appended to raster source IDs, instead of replacing `raster`


annehaley (approved):
Looks good. I was able to confirm that I no longer get the error that I was getting (`Error: Layer string incompatible: 119.0.531.raster.140.bounds.140.mask` when adding a raster layer)

PR #178, opened 2025-07-29
Switch from Webpack to Vite

### LINKED ISSUE #165: Switch to vite build tool
We should update our build tool for the client: vite is the more modern standard for building vue projects (instead of vue-cli-service)

jjnesbitt:
Closes #165


docs/setup.md
```diff
@@ -2,11 +2,9 @@
 
 1. Prepare the web client with the following steps:
 
-   a. Run `cp dev/.env.web.example dev/.env.web`.
-
-   b. Edit `dev/.env.web` and supply a valid MapTiler API key. To sign up for a free MapTiler API key, visit https://docs.maptiler.com/cloud/api/authentication-key/.
-
-   c. Install client dependencies by running `cd web && npm i`.
+   1. Run `cp web/.env.development web/.env`.
```
annehaley, line 5: Given this change, you can also delete `dev/.env.web.example`

web/components.d.ts
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable */
+// @ts-nocheck
+// Generated by unplugin-vue-components
+// Read more: https://github.com/vuejs/core/pull/3399
+// biome-ignore lint: disable
+export {}
+
+/* prettier-ignore */
+declare module 'vue' {
+  export interface GlobalComponents {
```
annehaley, line 10: If we create a new component and forget to add it to this list, would there be an informative error? If not, can we add some precommit hook or test to ensure that all existing components are present in this list?

annehaley (commented):
When I checked out this branch, I followed the new instructions to create `web/.env`, supply my MapTiler key, and install dependencies as instructed. When I bring up my containers, the web container prints the following expected output: 
```
web-1       | > uvdat@1.2.0 serve
web-1       | > vite
web-1       |   VITE v7.0.6  ready in 222 ms
web-1       |   ➜  Local:   http://localhost:8080/
web-1       |   ➜  Network: use --host to expose
```
However, when I go to localhost:8080 in my browser, it doesn't connect to anything. I still see that the docker compose config maps port 8080, so I'm not sure why this is happening.

web/components.d.ts
jjnesbitt: I think it will be auto-generated by vite, as this file was. So if we were to add a new component, this file would be automatically modified.

jjnesbitt:
> When I checked out this branch, I followed the new instructions to create `web/.env`, supply my MapTiler key, and install dependencies as instructed. When I bring up my containers, the web container prints the following expected output:
> 
> ```
> web-1       | > uvdat@1.2.0 serve
> web-1       | > vite
> web-1       |   VITE v7.0.6  ready in 222 ms
> web-1       |   ➜  Local:   http://localhost:8080/
> web-1       |   ➜  Network: use --host to expose
> ```
> 
> However, when I go to localhost:8080 in my browser, it doesn't connect to anything. I still see that the docker compose config maps port 8080, so I'm not sure why this is happening.

I had mainly tested this natively, and didn't realize it behaved differently I just pushed [6e81518](https://github.com/OpenGeoscience/uvdat/pull/178/commits/6e81518e850d0bda247969dac4ee1c3a825a8d11) which seems to work in both cases.

docs/setup.md
annehaley: I see that `web/.env.example` was removed, but not `dev/.env.web.example`

web/components.d.ts
annehaley: Ok great. I tested this out to confirm and it did automatically update.

web/src/components/sidebars/SideBars.vue
```diff
@@ -18,8 +18,8 @@ const panelStore = usePanelStore();
 const projectStore = useProjectStore();
 
 
-const version = process.env.VUE_APP_VERSION;
-const hash = process.env.VUE_APP_HASH;
+const version = import.meta.env.VITE_APP_VERSION;
+const hash = import.meta.env.VITE_APP_HASH;
```
annehaley, line 22: In all of the places that now use `import.meta.env`, I get the following typescript error: `Property 'env' does not exist on type 'ImportMeta'.` This only appears in my vscode, not in my browser when I use the app. 

I notice that in `vite.config.mts` you have `define: { 'process.env': {} },`, so maybe you could express everything from `import.meta.env` through `process.env` instead.

annehaley: Ok great, I confirmed that the ts error went away.

web/.env.example
```diff
@@ -1,4 +1,4 @@
-VUE_APP_OAUTH_CLIENT_ID=cBmD6D6F2YAmMWHNQZFPUr4OpaXVpW5w4Thod6Kj
-VUE_APP_API_ROOT=http://localhost:8000/api/v1
-VUE_APP_OAUTH_API_ROOT=http://localhost:8000/oauth/
-VUE_APP_MAPTILER_API_KEY=myMapTilerKey
+VITE_APP_OAUTH_CLIENT_ID=cBmD6D6F2YAmMWHNQZFPUr4OpaXVpW5w4Thod6Kj
+VITE_APP_API_ROOT=http://localhost:8000/
```
annehaley, line 2: Why was the `api/v1` part of this removed?

jjnesbitt: Because it was incorrect. [auth.js](https://github.com/OpenGeoscience/uvdat/blob/26693dfdd6f3227e7022d02d51ddad8e237db6ad/web/src/api/auth.js#L5) appends the `api/v1` to the end of that variable.

PR #183, opened 2025-08-07
New base tiles

annehaley:
This PR updates the base maps to use vector tiles served from `https://demo.kitware.com/vector-maps`. The following screenshots show the base map in light mode and dark mode.

image
image


web/src/store/map.ts
```diff
@@ -190,7 +193,7 @@ export const useMapStore = defineStore('map', () => {
   }
 
   function getUserMapLayers() {
-    return getMap().getLayersOrder().filter((id) => id !== 'base-tiles');
+    return getMap().getLayersOrder().filter((id) => id.includes('.'));
```
BryonLewis, line 196: I'm a bit wary about utilizing a string includes `'.'` to figure out what map layers are UserMapLayers.  You should know the source or the root name for all sources that are added to the map.  You sohuld be able to cross reference an array of your source names with the layer names.

Mainly, I'm worried that we may add some other base map sources at some point, or a style could change on the server and it could include a `'.'` at some point and then some random layer in a vector base map is included in your getUserLayers.

web/src/store/map.ts
```diff
@@ -164,12 +164,15 @@ export const useMapStore = defineStore('map', () => {
...
-      );
+    const dataLayers = getUserMapLayers()
+    map.getLayersOrder().forEach((id) => {
+      if (!dataLayers.includes(id)) {
+        map.setLayoutProperty(
+          id,
+          "visibility",
+          showMapBaseLayer.value ? "visible" : "none"
+        );
+      }
```
BryonLewis, line 175: In the future you may have multiple base Maps, like National Agriculture Imagery Program (NAIP): A Free Tile server of ariel imagery for the United States.
The Source can be added using this URL: https://gis.apfo.usda.gov/arcgis/rest/services/NAIP/USDA_CONUS_PRIME/ImageServer/tile/{z}/{y}/{x}?blankTile=false

If this is the case the toggling by not being a userLayer wouldn't work.  It would be nice if we could get the 'source' for the pmtiles and use that to get a list of layers to toggle off.
The style having the source in it makes it a little difficult but I think you should still be able to get the source name without magically knowing about it outside of the application.
You could use `map.getStyle().sources` which returns a dictionary of the source name with the type and url:
```
{
    "openstreetmap-openmaptiles":{
        "type": "vector",
        "url": "https://demo.kitware.com/vector-maps/tiles/planet-z15"
    }
}
You know that you are setting a mapStyle that is hosted at https://demo.kitware.com so you could search for a URL in the sources that begins with that to get the source name and use that source name to filter the Layers for those connected to the source.
```
const layersForSource = map.getStyle().layers.filter(
  (layer) => layer.source === sourceName
);
```

File: web/src/components/map/Map.vue [hunk 3]
```diff
@@ -60,49 +53,26 @@ function setAttributionControlStyle() {
   });
 }
 
+function setMapStyle() {
+  const map = mapStore.getMap();
+  map.setStyle(THEMES[appStore.theme].mapStyle)
```
BryonLewis, line 58:
```suggestion
  map.setStyle(THEMES[appStore.theme].mapStyle);
```

web/src/components/map/Map.vue
```diff
@@ -114,6 +84,9 @@ function createMap() {
   newMap.on("load", () => {
     mapContainer.classList.remove("spinner");
   });
+  newMap.on('styledata', () => {
+    layerStore.updateLayersShown()
```
BryonLewis, line 88:
```suggestion
    layerStore.updateLayersShown();
```

web/src/components/map/Map.vue
```diff
@@ -158,19 +131,8 @@ onMounted(() => {
...
-    type: "raster",
-    source: value,
-    minzoom: 0,
-    maxzoom: 22 + 1,
-    layout: {
-      visibility: mapStore.showMapBaseLayer ? "visible" : "none",
-    },
-  });
+watch(() => appStore.theme, () => {
+  setMapStyle()
```
BryonLewis, line 135:
```suggestion
  setMapStyle();
```

web/src/store/map.ts
```diff
@@ -164,12 +164,15 @@ export const useMapStore = defineStore('map', () => {
   // Update the base layer visibility
   watch(showMapBaseLayer, () => {
     const map = getMap();
-    getUserMapLayers().forEach((id) => {
-      map.setLayoutProperty(
-        id,
-        "visibility",
-        showMapBaseLayer.value ? "visible" : "none"
-      );
+    const dataLayers = getUserMapLayers()
```
BryonLewis, line 167:
```suggestion
    const dataLayers = getUserMapLayers();
```

BryonLewis (changes requested):
It functions exactly like I would expect it to.

Some minor ';' injections.

My comments are related to being more explicit when filtering and toggling layers.
You can leave it as is or push back against these changes if you feel the app would never hit these corner cases:
1.  where another layer has '.' in it that isn't a user layer 
2.  if you don't think you would ever have another base map layer

web/src/store/map.ts
annehaley: Refactored `getUserMapLayers` in [2272044](https://github.com/OpenGeoscience/uvdat/pull/183/commits/2272044f2141f9bb058cb36b66d7ad3800d667d8)

web/src/store/map.ts
annehaley: Added semicolons in [0ee84dc](https://github.com/OpenGeoscience/uvdat/pull/183/commits/0ee84dcae2c0b21dafbd82c9eccfe10d93e2431c)

web/src/store/map.ts
BryonLewis: A quick look at it makes me think it does.  The only possible downside to this is if we start hosting multiple base map layers at demo.kitware.com, but at that point I think we just extend the url further to include part of the path (not something we need to worry about right now).

BryonLewis (approved):
:+1 looks good

PR #186, opened 2025-08-26
Single JSON file Ingestion

### LINKED ISSUE #184: Populate/Ingest Data through Simplified JSON file
I think it may be nice if the populate script or another script would directly load a single JSON file for projects and raster/vector datasets.  To add a new project/dataset pair in the system now requires:

- Creating the containing folder
- ingest.py file that is loaded dynamically
- projects.json
- dataset.json
- Then you need to modify ingest_use_case.py to know about the new  folder/location

For simple uses cases having a single JSON file which is an array of Objects that can be of type 'project' or 'dataset'.  The system could then read the objects into memory and load the datasets first and then the projects:
i.e.
```
[
  {
    "name": "Sequential Test Rasters",
    "type": "dataset",
    "description": "Clipped Sentinel-2 images downloaded and clipped around point",
    "category": "imagery",
    "files": [
      {
        "path": "sequentialTestRasters/S2C_T18TWN_20250802T155417_L2A_visual_clip_10km.tif",
        "name": "Frame 0"
      },
      {
        "path": "sequentialTestRasters/S2C_T18TWN_20250723T155909_L2A_visual_clip_10km.tif",
        "name": "Frame 1"
      }
    ]
  },
  {
    "name": "SequentialTestRasters",
    "type": "project",
    "default_map_center": [42.4, -71.1],
    "default_map_zoom": 11,
    "datasets": ["Sequential Test Rasters"]
  }
]
```

BryonLewis:
resolves #184 

- Creates a new management command `ingest` that requires a JSON file as input
   - Options:
       - `--replace` - will replace all datasets and items in the Database by deleting any existing ones with the same name
       - `--clear` - prompts the user and then will clear all objects of Dataset, Chart or Project type. (I didn't double check the cascading but I believe this should remove the fileItems as well)
       - `--no-cache` - This will replace any downloaded file so that it grabs a new one and loads it in
- Format:  Single Json file that has an Array of Objects where each object has a type property that corresponds to `'Project' | 'Dataset' | 'Chart'`
- Attempted to add some typing and TypedDicts, they aren't used through pydantic or anything but I think it makes reasoning a bit easier about the structure of the data files
- I've changed the ingestion logic so that if a file exists it won't run the conversion process on the file again.  So multiple runs of the same file without a `--replace`, `--no-cache` or `--clear` will just speed through the ingestion by noticing all items are already there.
- Datasets have a new optional key `action` that can be `redownload` or `replace`.  This allows a single dataset in a json file to be replaced each time it is run.  This is extremely useful when testing ingestion/conversion scripts for a new datatype or file, or loading in files repeatedly.
- Datasets have another optional key `conversionScript` that is a relative path to a python file that containes `convert_dataset` that takes in a Dataset Django Model object and a Dataset JSON object from the file and runs a process on it.  This new format should dynamically import the file and run the conversion script so no additional imports are required.
- If the `conversionScript` is empty, it falls back on a default conversion process that simply spawns the `dataset.spawn_conversion_task`
- Added an environment Variable called `INGEST_BIND_MOUNT_POINT`.  Instead of always relying on the ./sample_data folder this could be an absolute path to another location for dataset ingestion.  This is handy on deployments where the user has access to the docker image but not the underlying code.  They can specify an environment variable with a full path to another location and bind mount it for data ingestion.
- Removed the previous `populate` management command
- Restructures the sample_data folder so there are top-level json files for boston_floods.json, la_wildfires.json, multiframe_test.json and there is a folder for new_york_energy because of it's custom ingestion mechanism.
- I removed the default ingest.py for systems that are using a simple dataset.spawn_conversion_process so the only one left is for new_york_energy.
- Updated documentation to reference the new ingest command.
- For testing I replaced calling the direct datasets with configurations options by adding two new json files `./sample_data/tests/ingest.json` and `./sample_data/tests/analytics.json` that should be the same as if loading those files with the include-large and dataset indexes.


sample_data/tests/ingest.json
```diff
@@ -0,0 +1,304 @@
+[
```
annehaley, line 1: The only objects we need to create for the ingest test are MBTA Rapid Transit (to test network evaluation), Massachusetts Elevation Data (raster evaluation), Boston Neighborhoods (regions evaluation), Boston Projected Flood Events (multilayer/multiframe evaluation), and any one Chart object. We can create a project too (we don't use the fixture in this test), but it should only contain those 4 datasets plus the chart.

uvdat/core/tests/test_ingest.py
```diff
@@ -30,13 +30,10 @@ def test_populate():
     # 4 is Massachusetts Elevation Data, tests raster eval
     # 6 is Boston Neighborhoods, tests regions eval
     # 9 is Boston Projected Flood Events, tests multilayer/multiframe dataset eval
-    dataset_indexes = [0, 4, 6, 9]
```
annehaley, line 33: You can also remove the comment explaining that `[0, 4, 6, 9]` are indices of target datasets

sample_data/boston_floods.json
```diff
@@ -1,5 +1,40 @@
...
+            "Massachusetts Elevation Data",
+            "Boston Orthoimagery",
+            "Boston Hurricane Surge Inundation Zones",
+            "Bsoton FEMA National Flood Hazard",
+            "Boston Neighborhoods",
+            "Boston Census 2020 Block Groups",
+            "Boston Zip Codes",
+            "Boston Projected Flood Events",
+            "Boston Flood Timeseries"
+        ]
```
annehaley, line 22: If you think that now is an appropriate time for it, I have been meaning to remove some unused/outdated datasets from our sample data. Namely, `Boston Hurricane Surge Inundation Zones` and `Boston FEMA National Flood Hazard` are unused (we added them at the beginning of Phase I to test vector data visualization before we had more appropriate data), and `Boston Projected Flood Events` and `Boston Flood Timeseries` are outdated (we used them for previous versions of the flood simulation). Additionally, the `Boston Orthoimagery` dataset has one layer we do use (the precomposited one) and one we do not use (the many-files one), so I would like to remove the latter.

If you'd rather not include that change in this PR, I can make a follow-up.

sample_data/boston_floods.json
```diff
@@ -1,5 +1,40 @@
...
+        "name": "DC Transportation",
+        "default_map_center": [
+            38.9,
+            -77.1
+        ],
+        "default_map_zoom": 11,
+        "datasets": [
+            "DC Metro"
+        ]
+    },
```
annehaley, line 35: Perhaps we should isolate the DC project and dataset in their own ingest JSON. I know they were a part of the `boston_floods` use case, but they should probably be separated.

uvdat/core/management/commands/ingest.py
```diff
@@ -0,0 +1,387 @@
...
+                self.stdout.write(self.style.ERROR(f"Invalid item type: {item.get('type')}"))
+                continue
+
+            if item['type'] == 'Project':
+                projects.append(item)
+            elif item['type'] == 'Dataset':
+                datasets.append(item)
+            elif item['type'] == 'Chart':
+                charts.append(item)
+        # Ingeset the datasets and charts now:
```
annehaley, line 161: I'd say to fix the typo, but this comment probably isn't ncessary given the print statements that follow

uvdat/core/management/commands/ingest.py
```diff
@@ -0,0 +1,387 @@
...
+                    'default_map_zoom': project['default_map_zoom'],
+                },
+            )
+            if created:
+                self.stdout.write(
+                    self.style.SUCCESS(f'Project {project_for_setting.name} created.')
+                )
+
+                project_for_setting.datasets.set(
+                    Dataset.objects.filter(name__in=project['datasets'])
```
annehaley, line 245: Could we add a warning if the project's list of dataset names includes anything that does not match an existing dataset?

uvdat/core/management/commands/ingest.py
```diff
@@ -0,0 +1,387 @@
...
+                        no_cache=no_cache or dataset.get('action', False) == 'redownload',
+                    )
+                dataset_for_conversion = new_dataset
+
+                dataset_size_mb = dataset_for_conversion.get_size() >> 20
+                self.stdout.write(
+                    self.style.SUCCESS(
+                        f'\t\t Dataset {dataset_for_conversion.name} of size {dataset_size_mb} MB.'
+                    )
+                )
```
annehaley, line 365: Is this left over from the `--include_large` part of the populate script? Or is it used here somehow?

scripts/sentinelDownload/sentinel2Download.py
annehaley, line 1: Could you explain what this script is used for? Maybe we could put a markdown file in the scripts folder to explain what each script is used for and provide example usages.

scripts/sentinelDownload/sentinel2Download.py
```diff
@@ -0,0 +1,220 @@
...
+# requires-python = ">=3.8"
+# dependencies = [
+#     "click",
+#     "pystac-client",
+#     "requests",
+#     "shapely",
+#     "rasterio",
+#     "pyproj",
+# ]
+# ///
```
annehaley, line 11: Since this script exists in dedicated `sentinelDownload` folder, you could probably just make a `requirements.txt` or `setup.py` file to go in there, too.

sample_data/tests/ingest.json
annehaley: You may need to update the expected counts in the assertions below, too

docs/setup.md
```diff
@@ -15,7 +15,9 @@
 
    c. Run `docker compose run --rm django ./manage.py makeclient` to create a client Application object for authentication.
 
-   d. Run `docker compose run --rm django ./manage.py populate` to use sample data.
+   d. Run `docker compose run --rm django ./manage.py ingest {JSON file}` to use sample data.
+
+   e. The JSON file can either be `multiframe_test.json`, `boston_floods.json`, `la_wildfires.json` or `./new_york_energy/data.json`
```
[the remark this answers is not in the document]
BryonLewis: Done, didn't know the ordered list hiearachy so I assumed roman numerials.  I.E 
1. First is Numbers
    a. Next is Letters
        i. Third is Roman Numerals

Let me know if you want something different.

scripts/sentinelDownload/sentinel2Download.py
BryonLewis: This metadata at the top is from [PEP 723](https://peps.python.org/pep-0723/).  It's meant to be similar to a `pyproject.toml` and includes the requirements for running the script independently.  If you use `uv run --script sentinel2Download.py`, it will automatically create a venv and install the dependencies. 

I think at some point UVDat as a whole should be updated to us `uv`, which is in line with the updated resonant stack.

docs/setup.md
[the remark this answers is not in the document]
BryonLewis: Looking at the preview of the Markdown file it looks like I can't do that because it thinks the i is part of the parrent list.  Using that format it looks like this:
image
It gives it this highlighted section instead of properlyt indenting.
If I change it to a number it properly indents:
image

You can mess with the example below in GithubMarkdown and see the difference

1. Prepare the web client with the following steps:

   1. Run `cp web/.env.example web/.env`.
   2. Install client dependencies by running `cd web && npm i`.

2. Run the docker containers with `docker compose up`. Be sure to check that all containers were able to start and stay running successfully before continuing.
3. While the containers are up, run the following commands in a separate terminal to prepare the database:

   a. Run `docker compose run --rm django ./manage.py migrate`.

   b. Run `docker compose run --rm django ./manage.py createsuperuser`
   and follow the prompts to create your own user.

   c. Run `docker compose run --rm django ./manage.py makeclient` to create a client Application object for authentication.

   d. Run `docker compose run --rm django ./manage.py ingest {JSON file}` to use sample data.

      - The JSON file can either be `multiframe_test.json`, `boston_floods.json`, `la_wildfires.json` or `./new_york_energy/data.json`

uvdat/core/management/commands/ingest.py
BryonLewis: https://github.com/OpenGeoscience/uvdat/blob/simplified-ingest/uvdat/core/management/commands/ingest.py#L244-L254

sample_data/tests/analytics.json
```diff
@@ -0,0 +1,260 @@
+[
```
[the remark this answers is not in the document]
BryonLewis: I added it back in and did a test ingest.

PR #192, opened 2025-09-22
Split large dataset upload test

annehaley:
This PR is made in response to [this comment](https://github.com/OpenGeoscience/uvdat/pull/191#discussion_r2363735396). This splits the large `test_rest_dataset_upload` function into three smaller tests.

### DIFF
--- a/uvdat/core/tests/test_dataset.py
+++ b/uvdat/core/tests/test_dataset.py
@@ -1,4 +1,4 @@
-from django.core.files.base import ContentFile
+from django.core.files.base import File
 import pytest
 
 from uvdat.core.models.project import Dataset
@@ -80,14 +80,17 @@ def test_rest_create_dataset(authenticated_api_client):
 
 @pytest.mark.django_db
 def test_rest_convert_dataset(
-    file_item, multiframe_vector_file, authenticated_api_client, project, user
+    file_item_factory, multiframe_vector_file, authenticated_api_client, project, user
 ):
+    with open(multiframe_vector_file['path'], 'rb') as f:
+        file_item = file_item_factory(
+            file=File(f),
+            name=multiframe_vector_file['name'],
+            file_type=multiframe_vector_file['file_type'],
+        )
     dataset = file_item.dataset
     project.set_collaborators([user])
     project.datasets.set([dataset])
-    with open(multiframe_vector_file['path'], 'rb') as f:
-        file_item.file.save(multiframe_vector_file['name'], ContentFile(f.read()))
-    file_item.file_type = multiframe_vector_file['file_type']
 
     result_expected = dict(
         name=f'Conversion of Dataset {dataset.name}',


uvdat/core/tests/conftest.py
```diff
@@ -80,6 +95,16 @@ def test_project() -> Project:
...
 
 
+@pytest.fixture
+def test_dataset(user, test_project) -> Dataset:
+    dataset = Dataset.objects.create(
+        name='Test Dataset', description='My test dataset', category='test'
+    )
+    test_project.set_collaborators([user])
+    test_project.datasets.set([dataset])
+    return dataset
```
jjnesbitt, line 105: This _could_ also be done using factories, but the use case is sort of split. Right now, we already have a `DatasetFactory`, which doesn't create a project by default. I think invoking `DatasetFactory` should create an associated `Project` by default (using the factoryboy [`post_generation`](https://factoryboy.readthedocs.io/en/stable/recipes.html#simple-many-to-many-relationship) hook).

Separate from that, in this case, you also want to add a collaborator to that project. This could maybe be considered a default in the `ProjectFactory` itself, maybe not. 

So if both behaviors can be considered default, then those factories could just be updates, and the `dataset` could simply be used. If not, I'd suggest naming this fixture something more fitting, like `dataset_with_collaborator` or something.

jjnesbitt (changes requested):

uvdat/core/tests/conftest.py
```diff
@@ -106,3 +131,41 @@ def permissions_client(user_info, test_project) -> APIClient:
...
+    resp = authenticated_api_client.post(
+        '/api/v1/files/',
+        dict(
+            name=TEST_FILE['name'],
+            file=s3ff_value,
+            file_type=TEST_FILE['file_type'],
+            dataset=test_dataset.id,
+        ),
+    )
+    return resp.json()
```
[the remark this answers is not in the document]
annehaley: Created `FileItemFactory` and associated fixtures in [8265989](https://github.com/OpenGeoscience/uvdat/pull/192/commits/8265989419231e58d8a333ae336bb8f0c28fc3e0). Part of the `file_item` fixture includes assigning the file field; I couldn't find a way to do it within the factory definition because it requires another fixture's value.

uvdat/core/tests/conftest.py
annehaley: Some other tests which focus on assigning datasets to projects or assigning project permissions to users fail if the `project` and `dataset` fixtures already include these assignments. We can't assume these assignments to be defaults, so I don't think we can include them in the factory definition. 

The `test_dataset` and `test_project` fixtures were intended to do those assignments for convenience, but they aren't really necessary. It's acceptable to do those assignments in the tests that need them, so I just removed them in [8265989](https://github.com/OpenGeoscience/uvdat/pull/192/commits/8265989419231e58d8a333ae336bb8f0c28fc3e0). Lmk your thoughts on this solution.

jjnesbitt: If the best solution for now is to simply set the dataset and collaborators in the test itself, then I'm fine with that.

PR #193, opened 2025-09-23
Colormap model

annehaley:
This PR creates a new model, `Colormap`. While #190 allows a user to create custom colormaps, those colormaps were limited by the fact that they were stored within a `LayerStyle` object's JSON field. This PR stores colormap definitions in their own table. `LayerStyle` objects simply reference colormaps by ID. 

A user can now edit and delete colormaps. When a colormap is deleted, any `LayerStyle` objects referencing that colormap will revert to using a default colormap.

Default colormaps are automatically created upon migration; these colormaps are not associated with any project. Custom colormaps created by users are associated with a project and can be edited/deleted by users with write permission on the project.


web/src/store/style.ts
```diff
@@ -57,31 +30,81 @@ interface NetworkStyle {
...
+    const colorB = markerB.color
+    const rgbA = {
+        r: parseInt(colorA.substring(1, 3), 16),
+        g: parseInt(colorA.substring(3, 5), 16),
+        b: parseInt(colorA.substring(5, 7), 16),
+    }
+    const rgbB = {
+        r: parseInt(colorB.substring(1, 3), 16),
+        g: parseInt(colorB.substring(3, 5), 16),
+        b: parseInt(colorB.substring(5, 7), 16),
```
zachmullen, line 47: Let's reduce the code duplication between this and the above block using a helper method like

```javascript
const toRgb = (hex) => {
  const num = parseInt(hex.substring(1), 16);
  return {
    r: (num >> 16) & 255,
    g: (num >> 8) & 255,
    b: num & 255,
  };
};
```
Then reuse it in both places.

uvdat/core/migrations/0014_colormaps.py
```diff
@@ -0,0 +1,60 @@
...
+                    models.ForeignKey(
+                        null=True,
+                        on_delete=django.db.models.deletion.CASCADE,
+                        related_name='colormaps',
+                        to='core.project',
+                    ),
+                ),
+            ],
+        ),
+        migrations.RunPython(forwards, backwards),
```
zachmullen, line 59: FWIW Django provides [RunPython.noop](https://docs.djangoproject.com/en/5.2/ref/migration-operations/#django.db.migrations.operations.RunPython.noop) that you can use as the value of `reverse_code` so that you don't have to define your own no-op method.

zachmullen (changes requested):

web/src/store/style.ts
```diff
@@ -57,31 +30,81 @@ interface NetworkStyle {
...
+    }
+    const rgbMid = {
+        r: Math.round((rgbA.r + rgbB.r) / 2),
+        g: Math.round((rgbA.g + rgbB.g) / 2),
+        b: Math.round((rgbA.b + rgbB.b) / 2),
+    }
```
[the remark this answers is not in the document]
annehaley: I went with `chroma-js` because its [`mix`](https://gka.github.io/chroma.js/#color-mix) function allows us to specify which color space to use. Changed in [9b91d93](https://github.com/OpenGeoscience/uvdat/pull/193/commits/9b91d933b812abe2b75263e5a904e6501dc73113)

web/src/store/style.ts
```diff
@@ -24,30 +25,9 @@ function getMidMarker(
...
-        r: Math.round((rgbA.r + rgbB.r) / 2),
-        g: Math.round((rgbA.g + rgbB.g) / 2),
-        b: Math.round((rgbA.b + rgbB.b) / 2),
-    }
-    const hexComponents = [
-        rgbMid.r.toString(16),
-        rgbMid.g.toString(16),
-        rgbMid.b.toString(16),
-    ].map((c) => c.length === 1 ? '0' + c : c)
+    const mid = chroma(markerA.color).mix(markerB.color, 0.5, 'lab')
```
zachmullen, line 28: ❤️

zachmullen (approved):

Issue #188, opened 2025-09-08
Switch to `uv`


annehaley:
This metadata at the top is from [PEP 723](https://peps.python.org/pep-0723/).  It's meant to be similar to a `pyproject.toml` and includes the requirements for running the script independently.  If you use `uv run --script sentinel2Download.py`, it will automatically create a venv and install the dependencies. 

I think at some point UVDat as a whole should be updated to us `uv`, which is in line with the updated resonant stack.

_Originally posted by @BryonLewis in https://github.com/OpenGeoscience/uvdat/pull/186#discussion_r2314250296_

PR #198, opened 2025-09-24
Fix CI failures

annehaley:
Our nightly CI tests have been failing since July 25 due to exceeding the maximum size allowed for CI containers. This PR changes three tests such that they now deal with smaller data to avoid hitting the max size limit.

1. `test_load_roads`: Simply use a smaller town. Pass in "Clifton Park" to the task instead of "Boston"
2. `test_ingest`: Use a smaller dataset for testing multiframe behavior
3. `test_flood_analysis_chain`: Use a new test flood data product that contains only 2 timesteps instead of 100

Previous failure: https://github.com/OpenGeoscience/uvdat/actions/runs/17964646245/job/51094609330
New success: https://github.com/OpenGeoscience/uvdat/actions/runs/17986517584/job/51165950002

Edit 1: Our `test` and `check-migrations` jobs began failing due to exceeding the max size, too.  [a283b56](https://github.com/OpenGeoscience/uvdat/pull/198/commits/a283b56682770b9cf37abe18d9d9e30579bccfb4) ensures that we don't install unnecessary dependencies, like prod deps and `geoai-py`.

Edit 2: The `bitnami/minio` image we use for CI is failing now, so we need to switch to using `bitnamilegacy/minio` like [this PR](https://github.com/kitware-resonant/cookiecutter-resonant/pull/373/files#diff-6b6e5067e5f2be7f76884a405b3180738fad9fc44f11f63fbb019641426e03fe) to the Resonant cookiecutter


setup.py
```diff
@@ -53,18 +53,13 @@
...
         'networkx==3.3',
         'pooch==1.8.2',
         'pyshp==2.3.1',
         'rasterio==1.3.10',
         'urllib3==1.26.15',
         'webcolors==24.6.0',
-        # Production-only
-        'django-composed-configuration[prod]==0.25.0',
-        'django-s3-file-field[boto3]==1.0.1',
-        'gunicorn==22.0.0',
```
jjnesbitt, line 67: Do you know that these are contributing towards the max size significantly? I think it's not ideal to have a `prod` install extra, and if we had a prod deployment, we might run into issues.

annehaley: Switching to `uv` has been on our list for a while now (#188), and in general I'd like to get uvdat more updated with resonant's upstream changes soon. @zachmullen I appreciate any advice or help you can offer in regards to those updates.

As for moving these dependencies to an extra, moving `geoai-py` alone was not enough to get under the size limit (see [this failure](https://github.com/OpenGeoscience/uvdat/actions/runs/17986746277/job/51333602619#step:7:62)), and the prod dependencies were the only other ones I could move without breaking the tests. They aren't big, but they were just enough to stay under the size limit. We have big dependencies (like GDAL) that we do need for the tests, and they get us pretty close to the limit.

jjnesbitt (approved):

PR #223, opened 2025-11-19
Add all requirements to `setup.py`

brianhelba:
Also add a `requirements.txt` file for Heroku

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+--find-links https://girder.github.io/large_image_wheels --editable .


setup.py
```diff
@@ -51,17 +51,20 @@
         'djangorestframework==3.15.2',
         'django-large-image==0.10.2',
         'drf-yasg==1.21.7',
+        'large-image[gdal]==1.33.3',
+        'large-image-converter==1.33.3',
         'matplotlib==3.9.2',  # for raster colormaps
         'geopandas==1.1.1',
         'networkx==3.3',
+        'numpy==2.2.6',
```
brianhelba: This was directly imported, but not listed.

PR #233, opened 2025-11-20
Fix some warnings

brianhelba:


setup.py
```diff
@@ -50,7 +50,7 @@
         'django-oauth-toolkit==2.4.0',
         'djangorestframework==3.15.2',
         'django-large-image==0.10.2',
-        'drf-yasg==1.21.7',
+        'drf-yasg==1.21.11',
```
brianhelba: Earlier versions give warnings when `setuptools` is not present at runtime.

PR #235, opened 2025-11-20
Update DB config for prod ASGI connections

annehaley:


brianhelba:
@annehaley @zachmullen I'm wondering whether we should also be using connection pooling in development? By adding `daphne` to `INSTALLED_APPS`, [`runserver` will run an ASGI development server](https://channels.readthedocs.io/en/latest/installation.html).

zachmullen:
Yes, in my previous ASGI setups that's what I did.

PR #239, opened 2025-11-21
Layer Comparison

### LINKED ISSUE #111: Advanced Interactions for Layer Comparison
Currently, the only way a user can compare two layers on the map is by toggling their visibility. We should add new interaction modes to the map that enable more advanced layer comparisons. This could include vertical/horizontal slides, circular/rectangular viewers that the user can resize, or other interaction modes. The scope of this issue includes research of existing solutions, determining which interactions are possible with MapLibre, and implementation. It may also include consulting Faiza on UX for the controls of these new interaction modes.

BryonLewis:
resolves #111 

- Adds `vue-maplibre-compare` dependency for comparing maps
- adds `useMapCompareStore` to handle the `isComparing` as well as properties for comparison
- Creates `MapWrapper.vue` to switch between default map and comparison map
- Adds a `mdi-compare` button to selected Layers to toggle into comparison mode
- Adds `CompareLayersPanel.vue` to panels that displays the left/right or top/bottom list of layers that can be toggled on/off for comparison.  This doesn't include the ability to reorder or change styles or remove layers.  To do that you need to go back to the main view

image

TODO:

- [x] Figure out why the maps have a 0 height when launched
- [x] Add the ouath.headers to the requests inside of the comparison library
    - [x] Need to update `vue-maplibre-compare` for that to be updated
- [x] Probably add emits for zoom center pitch and heading so when returning back to the map it is is in the same location.
    - [x] Confirm the emits are working
- [x] Fix Z-index issues with the compare map and the controls container
- [x] Decide on styling for the slider and coloring for dark/light modes
    - [x] Maybe look into using the theme''s primary color so it stands out
- [x] Look into using the left map as the base map and sliding over the swiper when it's enabled
- [x] Change the compare layer display to have dragging.  
- [x] Swap the mdi-cog action for a menu that would allow you to swap the currently displayed style
- [x] Look into trying to do this in a single list, where you have checkboxes for MapA and MapB and styles for MapA and MapB
- [x] Add support for removing/adding layers in the system
    - [x] Update the list when adding/removing so the ordering is true to the list
- [x] Toggle on/off the base map
- [x] Left/Right map having different color styling
    - [x] Style so that There is a single mdi-cog that enables tools to adjust left/right layer
    - [x] Make sure changing the color doesn't turn the layer on
    - [x] Vector map support
    - [x] Raster map support
    - [x] Opacity adjustment per left/right layer
- [x] Final Testing of reordering and any quirks with layering
- [x] Basemap toggling on/off and issues.

20251205 Updates/TODO:

- [x] Fix base Vector tile layer ordering issues.  When toggling sometimes the order of the items in the baseLayer seem to change causing some visual issues
    - [x] Expansion: the base map layer ordering if affected by the other ordering and it swaps back and forth.
- [x] Attribution text needs to be set for the `ToggleCompare`
- [x] Error when adding sources that already exists (specifically being in compare mode and attempting to add a new source)

Reviewer Questions:

- Keep Map.vue (original Map tool) and MapWrapper.vue (the reloading system) or only use the new ToggleCompare.vue
- Please do a quick overview of https://github.com/BryonLewis/vue-maplibre-compare and see if there aren any issues or modifications needed.

TODO 20251216:

- [x] MapB (second map) item selection and tooltip support
    - [x] Going to require adding all of the 'click' actions and events on the layers
    - [x] Update the centering of the map to take in account the slider.  Currently it attempts to center based on a full map and not the portion for MapA.  MapA and MapB need to be able to center based on their corresponding sizes.
        - [x] This will probably require an event stating the ratio of the comparison slider from inside of `vue-maplibre-compare` to know how to center the map
- [x] Basemaps don't change on MapB. Assuming this is some Async issue with style.json being requested and loaded. During testing it seems that the MapB basemap is one step behind the right one. I may have to do some async code when the basemap changes to handle this.
    - [x] When using a single URL for the style.json and loading it needs to process the style.json to get the layers and then load up the corresponding layers.  This is what is causing the async problem with changing map base layers while in comparison mode.
- [x] Add support for Horizontal and Vetical comparisons.
    - [x] Add maybe a hover state to the compare button that opens a v-menu allowing the user to swap between vertical and horizontal orientation
    - [x] Update all references to left/right to be top/bottom
    - [x] Makes sure that the reloading by changing the orientation works properly with everything
- [x] Reorient the CompareLayerStyle.vue
    - [x] Right/Left orientation for the vertical compare
    - [x] Top/Bottom orientation for the horizontal compare
- [x] Minor Issues
    - [x] Changing basemap will reset the styling to the MapA item
    - [x] Changing basemap will reset the opacity display
    - [x] Changing the style will reset the opacity display

### DIFF
--- a/web/src/components/map/MapTooltip.vue
+++ b/web/src/components/map/MapTooltip.vue
@@ -89,7 +89,6 @@ function zoomToFeature() {
   if (clickedFeature.value === undefined) {
     return;
   }
-
   // Set map zoom to match bounding box of region
   const map = mapStore.getMap(props.compareMap);
   const buffered = turf.buffer(
@@ -97,40 +96,12 @@ function zoomToFeature() {
     0.5, {units: 'kilometers'}
   )
   if (!buffered)  return;
-
   const bbox = turf.bbox(buffered);
   if (bbox.length !== 4) {
     throw new Error("Returned bbox should have 4 elements!");
   }
-  
-  // Handle padding based on orientation
-  if (compareStore.isComparing && compareStore.orientation === 'horizontal') {
-    // Horizontal orientation: top/bottom split
-    const padding = { top: 0, bottom: 0 };
-    let offset = window.innerHeight * compareStore.sliderEnd.percentage * 0.01;
-    // Offsets for the sidebars
-    if (!props.compareMap) {
-      padding.bottom = window.innerHeight - offset; // account for split
-      padding.top += appStore.openSidebars.includes("left") ? 0 : 0;
-    } else {
-      padding.top = offset; // account for split
-      padding.bottom += appStore.openSidebars.includes("right") ? 0 : 0;
-    }
-    map.fitBounds(bbox, {maxZoom: map.getZoom(), padding});
-  } else {
-    // Vertical orientation: left/right split (existing logic)
-    const padding = { left: 0, right: 0};
-    let offset = window.innerWidth * compareStore.sliderEnd.percentage * 0.01;
-    // Offsets for the sidebars
-    if (!props.compareMap) {
-      padding.right = window.innerWidth -offset; // account for sidebar
-      padding.left += appStore.openSidebars.includes("left") ? 300 : 0;
-    } else {
-      padding.left = offset; // account for sidebar
-      padding.right += appStore.openSidebars.includes("right") ? 300 : 0;
-    }
-    map.fitBounds(bbox, {maxZoom: map.getZoom(), padding});
-  }
+  map.fitBounds(bbox, {maxZoom: map.getZoom()});
+
 }
 
 // Check if the layer associated with the clicked feature is still selected and visible
@@ -162,7 +133,10 @@ watch(
     tooltip.setLngLat(center);
     // This makes the tooltip visible
     tooltip.addTo(mapStore.getMap(props.compareMap));
-    zoomToFeature()
+    // Don't zoom to feature if comparing maps
+    if (!compareStore.isComparing) {
+      zoomToFeature();
+    }
   }
 );
 


web/src/components/map/MapWrapper.vue
```diff
@@ -0,0 +1,120 @@
...
+            bearing: mapStats.value?.bearing,
+            pitch: mapStats.value?.pitch,
+        });
+    }
+});
+
+const mapStyleA = computed(() => compareStore.mapAStyle);
+const mapStyleB = computed(() => compareStore.mapBStyle);
+const mapLayersA = computed(() => compareStore.mapLayersA);
+const mapLayersB = computed(() => compareStore.mapLayersB);
```
annehaley: Why do you need to create computed versions of these store variables (as well as `computedCompare` and `mapStats`)?

web/src/components/map/MapWrapper.vue
```diff
@@ -0,0 +1,120 @@
...
+
+const swiperColor = computed(() => {
+    const theme = useTheme();
+    return theme.global.current.value.colors.primary
```
annehaley: When getting the theme like this, I don't think this respects when the user toggles the theme manually. If that's true, we can use `import { THEMES } from "@/themes";` and `THEMES[appStore.theme].colors.primary` instead.

web/src/store/compare.ts
```diff
@@ -0,0 +1,337 @@
...
+    pitch: number;
+}
+
+
+
+export const useMapCompareStore = defineStore('mapCompare', () => {
+    const mapStore = useMapStore();
+    const layerStore = useLayerStore();
+    const isComparing = ref<boolean>(false);
+    const orientation = ref<'horizontal' | 'vertical'>('vertical');
```
annehaley: It's good that we have this ref to be able to handle both modes, but I don't see anywhere in the UI to toggle its value.

web/src/components/map/ToggleCompareMap.vue
```diff
@@ -0,0 +1,266 @@
+<!-- eslint-disable vue/multi-word-component-names -->
+<script setup lang="ts">
+import { useAppStore, useLayerStore, useMapStore } from "@/store";
+import { useMapCompareStore } from "@/store/compare";
+import { computed, onMounted, Ref, ref, watch } from "vue";
```
annehaley: `onMounted` is an unused import

web/src/components/sidebars/CompareLayerStyle.vue
```diff
@@ -0,0 +1,361 @@
...
+                                    return-object
+                                    hide-details
+                                    @update:model-value="selectStyle($event, 'A')"
+                                ></v-select>
+                            </div>
+                            <table class="aligned-controls px-2">
+                                <tbody>
+                                    <tr v-if="frames.length > 1">
+                                        <td><v-label>Default Frame</v-label></td>
+                                        <td>
```
annehaley: I appreciate that you've kept the "default frame" input from the original styling menu, but I don't think it's applicable here. The "default frame" is a property of the saved style, and its input only makes sense if the style can be saved and updated. On multiframe layers, the layer panel (not the styling menu) will have a "current frame" slider.  A style's "default frame" is simply the default value for that "current frame" slider.

web/src/components/map/MapWrapper.vue
```diff
@@ -0,0 +1,120 @@
+<!-- eslint-disable vue/multi-word-component-names -->
```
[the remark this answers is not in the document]
BryonLewis: left over when copying over the Map.vue, will be removing this component anyways in this PR going forward to use the ToggleCompareMap.vue.  Will make sure that ToggleCompareMap.vue has the lint rule removed.

web/src/components/map/MapWrapper.vue
BryonLewis: Short answer: I do not strictly need the computed versions. I can update the code for consistency with the existing geoinsight style.  I will likely switch to Pinia’s [storeToRefs()](https://pinia.vuejs.org/api/pinia/functions/storeToRefs.html) so the reactivity of the store properties is explicit.

Longer Answer/Rant:

One of my goals is to minimize ambiguity around types and reactivity when reading code in the IDE. Vue’s `reactive()` creates objects that behave like plain objects at call sites, but the type information is implicit. When hovering, or when using autocompletion, a reactive object and a plain object look identical. For example:
```vue
const sample: { value: number } = { value: 10 };
const sample2: { value: number } = reactive({ value: 10});
const sampleRef: Ref<{value: number}> = ref(value: number);
```
In the IDE, both `sample` and `sample2` appear the same. Only sampleRef clearly indicates that it is reactive because .value is required for access. I have not been able to find a Volar/Vue Official or TS configuration that differentiates between a reactive object and a non-reactive one.

There could be some vue extenstion, tsconfig I'm missing but I've tried using Volar (now Vue Official) and different tsconfig settings and I've never had it indicate that a reactive variable is different from a standard one.

Using computed wrappers (and now I'll swap to `storeToRefs`) gives explicit clarity that the variables coming from the store are refs, not plain objects. This improves readability for me because I can immediately tell which values are reactive and which are simple local variables.

Now with the code above it's nice that `compareStore` indicates that all of the sub-values are Refs (the same thing that reactive does)

This is similar to why I use `defineComponent` a lot in my own Vue applications.  I want to explicitly indicate variables that are used in the template instead of assuming that all values are used.  There are times when variables are related to some store and used to derive the end state for the template and I like to quickly know what it is in the template and what isn't.

BryonLewis: swapped to `toStoreRefs` in https://github.com/OpenGeoscience/geoinsight/pull/239/commits/9278232fc9bbe3301f54882a6878954778624123

web/src/components/map/ToggleCompareMap.vue
```diff
@@ -0,0 +1,266 @@
...
+  customAttribution: ATTRIBUTION,
+});
+attributionControl.onAdd = (map: Map): HTMLElement => {
+  attributionControl._map = map;
+  const container = document.createElement("div");
+  container.innerHTML = ATTRIBUTION.join("");
+  attributionControl._container = container;
+  setAttributionControlStyle();
+  return container;
+};
```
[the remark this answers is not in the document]
BryonLewis: Map.vue is a leftover, it it isn't being used I'll remove it.

web/src/store/compare.ts
```diff
@@ -0,0 +1,337 @@
...
+import { MapLayerStyleRaw } from './style';
+import { LayerStyle } from '@/types';
+
+interface DisplayCompareMapLayerItem {
+    displayName: string;
+    state: boolean;
+    layerIds: string[];
+}
+
+interface DisplaycompareMapLayer {
```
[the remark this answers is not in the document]
BryonLewis: resolved in https://github.com/OpenGeoscience/geoinsight/pull/239/commits/198bbef1b343afd136056f274e4e783744d318d7

web/src/store/compare.ts
annehaley: Sure, why not? As long as it doesn't add too much complexity, it couldn't hurt to give the user the option. It seems like you have support for it already and we just need the button to toggle it.

web/src/components/map/ToggleCompareMap.vue
```diff
@@ -0,0 +1,266 @@
...
+    });
+
+    /**
+     * This is called on every click, and technically hides the tooltip on every click.
+     * However, if a feature layer is clicked, that event is fired after this one, and the
+     * tooltip is re-enabled and rendered with the desired contents. The net result is that
+     * this only has a real effect when the base map is clicked, as that means that no other
+     * feature layer can "catch" the event, and the tooltip stays hidden.
+     */
+    newMap.on("click", () => {mapStore.clickedFeature = undefined});
```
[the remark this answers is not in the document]
BryonLewis: Notes to myself:

- The left map works with clicking, it just centers the map when clicked which places the tooltip behind the compare if it is left at it's default location.
    - I probably need vue-maplibe-compare to return back the split location so that way I can compute a different centering location when clicking.  I.E I want it to center using the new width/height of the compare area and not the full map
- MapB:
    - This one is harder because the click handlers for each layer isn't added when the Style is copied over.  For this to work I need to extract all of the click handlers and make sure that when layers are added/removed that the click handlers also get added to those layers and removed properly when those layers are removed.
    - This would probably be done inside of `compare.ts` and the watcher on `selectedLayers` that will add/remove layers.  
    - It also needs to work when the compare is initially toggled on so that layers properly work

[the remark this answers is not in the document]
BryonLewis: Updated in https://github.com/OpenGeoscience/geoinsight/pull/239/commits/2d9b0a52f736c241e07d15acc61f3184af18dd25

Small Demo video below of having the two tooltips.
https://github.com/user-attachments/assets/50d38316-643f-4900-87af-e37cc83959c3

web/src/store/compare.ts
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    const getBaseLayerSourceIds = () => {
+        const map = mapStore.getMap();
+        if (!map) return [];
+        return map.getStyle().layers.filter((layer: any) => {
+            const layerWithSource = layer as { source?: string };
+            return layerWithSource.source?.includes('openstreetmap');
+        }).map((layer: any) => layer.id);
+    }
```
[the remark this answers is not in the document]
BryonLewis: I've made a different function in map.ts `getBaseLayerIds`  in https://github.com/OpenGeoscience/geoinsight/pull/239/commits/f4f0a2a08243b51414011af782d0f2e85bb33497

web/src/store/compare.ts
BryonLewis: I don't know if I'm fully okay with the interface but what I have now is below
image
image

We may want to invert the comparison names.  Horizontal means that the slider is horizontal.  I could see where horizontal means comparing left/right.

Implemented in: https://github.com/OpenGeoscience/geoinsight/pull/239/commits/9a632c2b8c4426e6105f5ba66844bba32f94ca62

PR #243, opened 2025-11-28
Fix Docker build and runtime issues for cross-platform support

aashish24:
Build fixes:
- Upgrade rasterio to 1.3.11 (fixes numpy==2.0.0rc1 build dependency)
- Add g++ and libgdal-dev for building rasterio from source on ARM
- Set GDAL_CONFIG env var for Fiona/tile2net builds

Runtime fixes:
- Add optional GPU profile for NVIDIA systems (celery-gpu service)
- Default celery runs CPU-only, use --profile gpu for GPU acceleration
- Fix web container to install npm dependencies inside container
- Use named volume for node_modules to avoid host/container arch mismatch
- Update Node.js to v22 (required by Vite 7.x)
- Remove platform: linux/amd64 from postgres for native ARM performance

Documentation:
- Update setup.md with GPU instructions and troubleshooting
- Improve formatting and add prerequisites section

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -38,13 +38,18 @@ While the containers are running, open a **separate terminal** and run:
 docker compose run --rm django python manage.py migrate
 
 # Create an admin user (you will be prompted for email and password)
-docker compose run --rm django python manage.py createsuperuser
+docker compose run --rm -it django python manage.py createsuperuser
 
 # Create OAuth client for authentication
 docker compose run --rm django python manage.py makeclient
 ```
 
 > **Note:** The `createsuperuser` command prompts you to create login credentials (email and password). Use these credentials to sign into both the Admin Panel and User Interface. If you forget your password, run `createsuperuser` again to create a new admin account.
+
+> **Windows Users:** If the `createsuperuser` command hangs or doesn't show prompts, prefix the command with `winpty`:
+> ```bash
+> winpty docker compose run --rm -it django python manage.py createsuperuser
+> ```
 
 ### 4. Load Sample Data (Optional)
 
@@ -157,3 +162,23 @@ docker compose up
 ```
 
 GPU acceleration is optional and only needed for accelerated inferencing.
+
+### Windows-Specific Issues
+
+**Interactive commands don't work or hang:**
+
+On Windows (especially Git Bash/MINGW), interactive Docker commands like `createsuperuser` may hang or fail to display prompts. Prefix the command with `winpty`:
+
+```bash
+winpty docker compose run --rm -it django python manage.py createsuperuser
+```
+
+**"No such container" errors:**
+
+If you see errors like `No such container: django` when trying to attach to a running container, use `docker compose run` instead:
+
+```bash
+docker compose run --rm -it django python manage.py <command>
+```
+
+This creates a new container instance rather than attaching to an existing one.


docs/setup.md
```diff
@@ -1,36 +1,156 @@
...
+```
+
+> **Note:** GPU mode requires NVIDIA drivers and [nvidia-docker](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html) runtime. This enables accelerated tile2net inference.
+
+### Access Points
+
+| Service | URL |
+|---------|-----|
+| Admin Panel | http://localhost:8000/admin/ |
+| User Interface | http://localhost:8080/ |
```
annehaley: I like this little table. We could also add `http://localhost:8000/api/docs/swagger/`, the API reference page.

docs/setup.md
```diff
@@ -1,36 +1,156 @@
...
+### 3. Initialize the Database
 
-1. Run `docker compose pull`
-2. Run `docker compose build --pull --no-cache`
-3. Run `docker compose run --rm django ./manage.py migrate`
+While the containers are running, open a **separate terminal** and run:
+
+```bash
+# Apply database migrations
+docker compose run --rm django ./manage.py migrate
```
annehaley: While working with August to get his local instance running, we learned that on Windows, invoking `./manage.py` directly encounters a carriage return issue. Invoking it with `python manage.py` instead works fine. Let's replace all of the `./manage.py` invocations in these instructions with `python manage.py`.

zachmullen:
I just did a fresh `docker compose build --no-cache` on my Apple Silicon machine and it worked without needing to set any platforms to `amd64`. Awesome!

docs/setup.md
augustposch: I got as far as seeing the Admin Panel login page and seeing the User Interface login page. As a next step, can someone assign me a login please?

docker-compose.override.yml
```diff
@@ -44,26 +45,61 @@ services:
     env_file: ./dev/.env.docker-compose
     volumes:
       - .:/opt/geoinsight-server
-    # NVIDIA driver and increased memory limit are required for tile2net inference
+    shm_size: 1gb
```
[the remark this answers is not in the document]
annehaley: After #248, we no longer require a GPU-enabled container for the `tie2net` task. We can still keep the option around for other accelerated inferencing, though.

docs/setup.md
```diff
@@ -1,36 +1,163 @@
...
-   d. Run `docker compose run --rm django ./manage.py ingest {JSON file}` to use sample data.
+```bash
+# Copy environment configuration
+cp web/.env.example web/.env
 
-      - The JSON file can either be `multiframe_test.json`, `boston_floods.json`, `la_wildfires.json` or `./new_york_energy/data.json`
+# Install frontend dependencies
+cd web && npm i
+cd ..
+```
```
annehaley: The installation of node modules is now included in the web container's command, so we can omit this section from the setup guide.

aashish24:
@annehaley thanks for your review. I addressed your comments. I kept reference to node installation in case developers might work on frontend outside containers

augustposch (changes requested):
image
This time `docker compose up` didn't work for me due to missing semicolon. Skimming the yaml files, I'm not sure why there would be a problem with a semicolon in line 48. Any suggestion for me?

PR #251, opened 2025-12-16
Basemap menu adjustments

annehaley:
This PR makes some minor changes to the appearance of the basemap menu:

- For mini previews of existing basemaps, define the map position with the _bounds_ of the main map instead of the center and zoom. This eliminates the cropping effect due to the smaller map container.
- Move mini previews to the left of basemap names
- Modify the aspect ratio of mini previews (landscape rectangle instead of square)

image


BryonLewis (approved):
:+1:

PR #254, opened 2025-12-18
fix: remove bounds vector layers with raster removal

### LINKED ISSUE #253: Bug:  Adding Removing and re-adding a raster dataset causes an error
Steps to Reproduce:

1.  Use the Boston Floods dataset
2. Add the Elevation data (I think any raster data)
3. Remove the Elevation data
4. Try to add the elevation data again

Image

Shows that it isn't removing the associated source for the vector data with the raster.
I.E. `removeLayers` in `map.ts` doesnt' account for the `id.bounds` vector source that is added to each layer

BryonLewis:
resolves #253 

When rasters are added to the map they add another `.bounds.` layer that is a vector layer.  When removing the raster the layer still exists and the source isn't removed.  So when the user removes a raster layer and then adds it again there would be an error.

- Added `.bounds.` to the filter condition when `getBaseLayerSourceIds()` so that way the bounds layer isn't included in the list
- Updated `removeLayers()` so that if the removal is of type `.raster.` it will remove the same layerId with `.bounds.` and remove the same source

### DIFF
--- a/web/src/store/map.ts
+++ b/web/src/store/map.ts
@@ -214,7 +214,7 @@ export const useMapStore = defineStore('map', () => {
   function getBaseLayerSourceIds() {
     return getMapSources()
       .filter((sourceId) => (
-        !sourceId || !(sourceId.includes('.vector.') || sourceId.includes('.raster.'))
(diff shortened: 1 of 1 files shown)


annehaley (approved):
LGTM, thanks for identifying and resolving this!

PR #260, opened 2025-12-30
Use a static asset for the Kitware logo

annehaley:
Our deployed instance at https://www.geoinsight.kitware.com/ encounters CORS errors when fetching the favicon and title logo at https://www.kitware.com/main/wp-content/uploads/2021/12/Kitware-Mark-231x300.png. This PR includes the image as a static asset in the webapp.


brianhelba (changes requested):
Right now, the asset will not get included in the build, since Vite doesn't know about the reference. You have two options:
* `import` the file to get a URL to it: https://vite.dev/guide/assets#importing-asset-as-url
* Use a `public` directory (following the advice about how to reference it): https://vite.dev/guide/assets#the-public-directory

Separately, the actual PNG image file here isn't rendering for me. I'd suggest pulling a fresh copy from [Kitware's assets in our company drive](https://drive.google.com/drive/folders/1tky7h9zlXNJeAZTD5IWMTvT0BQ7B7aSD).

PR #266, opened 2026-01-05
Refactor Analytics Websocket connection

annehaley:
@augustposch helped me discover that the analytics websocket connection is not created if the Analytics panel is not mounted when the current project is changed. This PR moves that logic from `AnalyticsPanel.vue` to `store/analysis.ts` so that the websocket creation is guaranteed whenever the current project is changed.


augustposch:
I switched to the new branch and tried a new flood simulation immediately, and that one didn't push the notification to the app, so I had to switch tabs back and forth to make it appear. Then I took down the docker containers and redid `docker compose up`, and then push worked.

PR #269, opened 2026-01-13
use asset importing and building for Kitware logo png

BryonLewis:
I understand having the favicon.ico i  public because index.html is a fairly static file.

Brian's suggestions for importing other assets and the tree shaking that building provides would be better than placing more assets into the public folder.

This PR moves the kitware.png file into src/assets and imports that directly so the build JS can handle the location of the file.

UPDATE:

- modified `.gitattributes` for some reason it was treating all files as text instead of binary.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,7 @@
 # Code needs to be run with Unix-style line endings, even when checked out on Windows
-* text eol=lf
+* text=auto eol=lf
 *.png binary
+*.jpg binary
+*.jpeg binary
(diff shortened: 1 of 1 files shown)


BryonLewis:
Sorry, I decided to make the .gitattributes a bit more explicit.

Hopefully the `text=auto` from https://git-scm.com/docs/gitattributes#_eol makes it so it doesn't try to convert other binaries to text.  I've added a few more image types to hopefully avoid this issue in the future.

PR #276, opened 2026-02-03
Only trigger Cloudflare Pages upon changes to `web/`

brianhelba:


brianhelba:
Looks like Terraform is unhappy. I'll investigate.

brianhelba:
Fixed in #405 now

PR #280, opened 2026-02-09
Add amenities dataset to boston floods ingest data

annehaley:
The added dataset has 15 layers (shown in screenshot below), each containing a dataset from https://maps.massgis.digital.mass.gov/MassMapper/MassMapper.html. 

image

Each layer contains point features to depict the locations of that amenity type. The screenshot below shows the "schools" layer:
Screenshot from 2026-02-09 14-54-16


BryonLewis (approved):
:+1

One small note.  I had already ingested the boston_flood data previously.  I would have thought that reingestion would add the data to the project.  While it added the new layer I needed to manually go in and add the Amenities layer to the project in the UI.  Small thing I'm not worried about but just wanted to let you know.

PR #287, opened 2026-02-16
MBTA ridership data

annehaley:
This PR leverages the `ingest_mbta` custom dataset conversion to add ridership data to the MBTA Rapid Transit dataset. Ridership data was provided by @augustposch. This PR also makes a slight change to the Legend panel so that the ridership range will appear on either side of the color bar.

**Note:** To apply these changes, the MBTA Rapid Transit dataset must be overwritten. Re-run the ingest task with "action": "replace" added to the JSON specification for that dataset.

### DIFF
--- a/sample_data/boston_floods/ingest_mbta.py
+++ b/sample_data/boston_floods/ingest_mbta.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
+
 import io
 
-import pandas
+import pandas as pd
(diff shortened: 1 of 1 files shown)


augustposch (approved):
New screenshot looks good. Glad you identified all the issues and fixed these! I'll look into the silver line and red line streetcar data this week.

PR #289, opened 2026-02-19
Upgrade Python to 3.12

annehaley:


brianhelba (commented):
LGTM, and it looks like `scikit-learn` is indeed at 1.8.0 now.

annehaley:
It took some hunting, but I did find a spot that needed a `with` block. I didn't see it at first because we don't explicitly invoke the `tempfile` library, but there was a spot where we open an S3FileField file without closing it. Fixed in [c1a5f53](https://github.com/OpenGeoscience/geoinsight/pull/289/commits/c1a5f5365a76f9ace7c0e078ec8f8aebedcae011).

PR #290, opened 2026-02-19
Add worker update script

annehaley:


terraform/update_worker.sh
```diff
@@ -0,0 +1,29 @@
...
+echo "==> Fetching secrets from Heroku..."
+ENV_FILE=$(mktemp)
+heroku config --app geoinsight --shell > "$ENV_FILE"
+
+echo "==> Uploading env file to remote machine..."
+scp "$ENV_FILE" "$SSH_TARGET:/home/ubuntu/geoinsight.prod.env"
+rm "$ENV_FILE"
+
+echo "==> Adding SOURCE_VERSION to env file..."
+ssh "$SSH_TARGET" 'echo "SOURCE_VERSION=$(git -C /home/ubuntu/geoinsight rev-parse HEAD)" >> /home/ubuntu/geoinsight.prod.env'
```
brianhelba: This should come after `git pull`, since we want to inform the application of the SHA of the the Git commit that it'll actually be running after this update.

This variable is ultimately used to feed metadata of Sentry errors.

PR #294, opened 2026-02-23
Replace `filter_queryset_by_projects` with a `ProjectQuerySet` class

brianhelba:


geoinsight/core/rest/access_control.py
```diff
@@ -47,15 +45,13 @@ def has_object_permission(self, request, view, obj):
...
+        if not hasattr(queryset, "filter_by_projects"):
             raise NotImplementedError
 
         ids = request.query_params.get("project", request.query_params.get("project_id"))
         if ids:
             # Return queryset filtered by objects that are within these projects
-            return queryset.model.filter_queryset_by_projects(
-                queryset=queryset, projects=Project.objects.filter(id__in=ids)
-            )
+            return queryset.filter_by_projects(Project.objects.filter(id__in=ids))
```
annehaley: If the request specifies the project IDs, then we shouldn't return objects with a null project relationship. Maybe this could be specified as a boolean argument to `filter_by_projects`?

annehaley (dismissed):

PR #296, opened 2026-02-24
Use Celery's `Task.apply` when immediately invoking tasks

brianhelba:
This ensures that the call is still processed by all of Celery's task machinery, improving parity with asynchronous invocation.


brianhelba:
See: https://docs.celeryq.dev/en/stable/userguide/canvas.html#partials

PR #303, opened 2026-02-26
Shareable Views

### LINKED ISSUE #73: Brainstorm "publishable maps" feature

annehaley:
Depends on #299.
Resolves #73.

This PR creates a new model `ViewState` and adds a menu for saving/loading ViewStates to the `ControlsBar` component.
image

When the user selects a view state from this list, the window will be redirected to `/view/{id}` and the view state will be loaded from the url. The user may refresh the page or send the link to a collaborator and the same UI state will be loaded. Therefore, the view states listed in the menu each have a "copy shareable link" button to the side.

When the user clicks the "Save Current View State" button, this dialog appears:
image

### DIFF
--- a/web/src/store/project.ts
+++ b/web/src/store/project.ts
@@ -251,7 +251,9 @@ export const useProjectStore = defineStore('project', () => {
     }
 
     async function refreshAllDatasets() {
-        allDatasets.value = await getDatasets()
+      loadingDatasets.value = true;
+      allDatasets.value = await getDatasets()
+      loadingDatasets.value = false;
     }
 
     watch(projectConfigMode, loadProjects);
--- a/web/src/components/projects/ProjectConfig.vue
+++ b/web/src/components/projects/ProjectConfig.vue
@@ -136,7 +136,6 @@ function saveProjectMapLocation(project: Project | undefined) {
 function selectProject(project: Project) {
   if (selectedProject.value?.id !== project.id) {
     selectedProject.value = project;
-    projectStore.loadingDatasets = true;
     projectStore.refreshAllDatasets()
     refreshProjectDatasets(null);
   }
@@ -222,12 +221,6 @@ onMounted(() => {
 });
 
 watch(selectedProject, resetProjectEdit);
-
-watch(() => projectStore.allDatasets, () => {
-  if (projectStore.allDatasets && projDatasets.value) {
-    projectStore.loadingDatasets = false;
-  }
-})
 
 watch(() => projectStore.projectConfigMode, () => {
   if (projectStore.currentProject) {


uvdat/core/rest/serializers.py
```diff
@@ -281,3 +282,24 @@ def get_name(self, obj):
...
+        # Web client expects Lon, Lat
+        if obj.map_center:
+            return [obj.map_center.y, obj.map_center.x]
+        return None
+
+    def to_internal_value(self, data):
+        center = data.get("map_center")
+        data = super().to_internal_value(data)
+        if isinstance(center, list):
+            data["map_center"] = Point(center[1], center[0])
```
BryonLewis: These should probably be reversed, in the admin panel these are reversed for the point viewer.

uvdat/urls.py
```diff
@@ -50,6 +51,7 @@
 router.register(r"networks", NetworkViewSet, basename="networks")
 router.register(r"basemaps", BasemapViewSet, basename="basemaps")
 router.register(r"analytics", AnalyticsViewSet, basename="analytics")
+router.register(r"views", ViewViewSet, basename="views")
```
BryonLewis: Similar to other comments, questioning if the shared map state should be labelled just views

web/src/components/map/ToggleCompareMap.vue
```diff
@@ -205,65 +206,50 @@ const updateBasemap = () => {
...
-            @pitchend="compareStore.updateMapStats($event)"
-            @rotateend="compareStore.updateMapStats($event)"
-            @sliderend="compareStore.updateSlider($event)"
-            @map-ready-a="handleMapReady($event, 'A')"
-            @map-ready-b="handleMapReady($event, 'B')"
-            class="map"
-        />
+  <div>
+    <ToggleCompare :map-style-a="mapStyleA" :map-style-b="mapBStyle" :map-layers-a="mapLayersA"
+      :map-layers-b="mapLayersB" :compare-enabled="compareStore.isComparing" :camera="{
```
BryonLewis: I'm guessing a formatter did this, was it intentional?

web/src/store/project.ts
```diff
@@ -50,17 +60,172 @@ export const useProjectStore = defineStore('project', () => {
         getDatasetTags().then((tags) => availableDatasetTags.value = tags)
     }
 
-    watch(currentProject, () => {
+        function fetchProjectViews() {
+        if (!currentProject.value) { return; }
+
+        getProjectViews(currentProject.value.id).then((views) => {
```
BryonLewis: async await?

uvdat/core/rest/view.py
```diff
@@ -0,0 +1,15 @@
...
+from uvdat.core.rest.access_control import GuardianFilter, GuardianPermission
+from uvdat.core.rest.serializers import ViewSerializer
+
+
+class ViewViewSet(ModelViewSet):
+    queryset = View.objects.all()
+    serializer_class = ViewSerializer
+    permission_classes = [GuardianPermission]
+    filter_backends = [GuardianFilter]
+    lookup_field = "id"
```
brianhelba: Remove these, in light of recent PRs?

web/src/store/project.ts
```diff
@@ -50,17 +60,172 @@ export const useProjectStore = defineStore('project', () => {
         getDatasetTags().then((tags) => availableDatasetTags.value = tags)
     }
 
-    watch(currentProject, () => {
+        function fetchProjectViews() {
```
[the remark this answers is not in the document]
annehaley: Changed in [db29318](https://github.com/OpenGeoscience/geodatalytics/pull/303/commits/db2931859e01e5e252fffaf1ab7466c8e32e9ef5)

uvdat/core/models/view.py
```diff
@@ -0,0 +1,58 @@
...
+        Basemap, on_delete=models.CASCADE, related_name="views", null=True
+    )
+    current_chart = models.ForeignKey(
+        Chart, on_delete=models.CASCADE, related_name="views", null=True
+    )
+    current_network = models.ForeignKey(
+        Network, on_delete=models.CASCADE, related_name="views", null=True
+    )
+    left_sidebar_open = models.BooleanField(default=False)
+    right_sidebar_open = models.BooleanField(default=False)
```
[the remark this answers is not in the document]
annehaley: They could go in the panel arrangement, but they are not technically part of that store ref in the client. The `app.ts` store has a ref `openSidebars` with the following type: `ref<("left" | "right")[]>`. 

I was going to use another JSONField to store this in the database, but I figured that it's better to use basic field types when possible (and only use a JSONField when necessary).

BryonLewis (changes requested):
There is another issue I believe with the Projects configuration page and specifically the ProjectDataset.vue where it isn't properly setting the `projectStore.loadingDatasets` to false when you have a project selected and go to that page.

PR #306, opened 2026-02-27
Complete logouts with allauth logout page

### LINKED ISSUE #300: Incomplete logout
When a user clicks the "logout" button in the web client, they aren't fully logged out. Clicking the "login" button afterwards will complete authentication with one click rather than bringing the user to the sign-in page to provide credentials.

annehaley:
Resolves #300

### DIFF
--- a/web/src/api/auth.js
+++ b/web/src/api/auth.js
@@ -1,7 +1,7 @@
 import axios from "axios";
 import OauthClient from "@resonant/oauth-client";
 import S3FileFieldClient from 'django-s3-file-field';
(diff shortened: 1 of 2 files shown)


brianhelba (approved):
There is another behavior change here, which is that the client-level logout is now unconditional (instead of being behind a confirm dialog). It'd be possible to logout from the client, then decline to logout from the Django server.

I think this is fine and probably still better than preserving the existing confirm box, which would require the user to confirm twice (once on the client, once on the Django server).

PR #386, opened 2026-05-13
Misc client bug fixes

annehaley:
This PR addresses various client bugs (some of which were caught by Sentry in production, some of which I found during the process).

Fixed Bugs
- Sentry caught this error when a request was unsuccessful: "Cannot read properties of undefined (reading 'results')". This can be solved with the "?" optional chaining operator wherever we access `.data.results` in `rest.ts` 
- The previews in the basemap options menu were broken due to the format of the bounds value. I don't know what changed to break this since I implemented it, but converting the format of the bounds fixed it.
- The screenshot feature was broken, raising the following error: "Unsupported color format color()". I swapped out `html2canvas` for [`html2canvas-pro`](https://www.npmjs.com/package/html2canvas-pro), which is a fork of the original with advanced color support. Again, I don't know what changed since I implemented screenshots, but the dependency swap fixes it.
- In the basemap creation dialog on the "MapLibre Style JSON" tab, the instructions state that you can paste JSON or you can specify a URL. However, the validation was broken if you pasted a URL; it tried to validate that string instead of the content at the url. I added a fetch statement wrapped in a try-except to handle this.

Other Improvements
- I fixed a few styling things that I noticed: some icons lacked colors and the spacing was too compact around the project config section of the left sidebar
- The menu to toggle the visibility of panels was a bit frustrating; you could click the name of a panel and it would animate like the click did something, but the toggle would only apply if you clicked directly on the checkbox. I fixed this so that clicking anywhere on the list item would apply the toggle.

### DIFF
--- a/web/src/api/rest.ts
+++ b/web/src/api/rest.ts
@@ -25,11 +25,11 @@ import type {
 } from "@/types";
 
 export async function getUsers(): Promise<User[]> {
-  return (await apiClient.get("users/")).data.results;
+  return (await apiClient.get("users/")).data?.results;
 }
 
 export async function getBasemaps(): Promise<Basemap[]> {
-  return (await apiClient.get("basemaps/")).data.results;
+  return (await apiClient.get("basemaps/")).data?.results;
 }
 
 export async function createBasemap(basemap: Basemap): Promise<Basemap> {
@@ -37,7 +37,7 @@ export async function createBasemap(basemap: Basemap): Promise<Basemap> {
 }
 
 export async function getProjects(): Promise<Project[]> {
-  return (await apiClient.get("projects/")).data.results;
+  return (await apiClient.get("projects/")).data?.results;
 }
 
 export async function createProject(
@@ -75,7 +75,7 @@ export async function deleteProject(projectId: number): Promise<Project> {
 export async function getProjectDatasets(
   projectId: number,
 ): Promise<Dataset[]> {
-  return (await apiClient.get(`datasets/?project=${projectId}`)).data.results;
+  return (await apiClient.get(`datasets/?project=${projectId}`)).data?.results;
 }
 
 export async function getChart(chartId: number): Promise<Chart> {
@@ -87,7 +87,7 @@ export async function getChartFiles(chartId: number): Promise<FileItem[]> {
 }
 
 export async function getProjectCharts(projectId: number): Promise<Chart[]> {
-  return (await apiClient.get(`charts/?project=${projectId}`)).data.results;
+  return (await apiClient.get(`charts/?project=${projectId}`)).data?.results;
 }
 
 export async function getProjectAnalysisTypes(
@@ -97,7 +97,7 @@ export async function getProjectAnalysisTypes(
 }
 
 export async function getDatasets(): Promise<Dataset[]> {
-  return (await apiClient.get("datasets/")).data.results;
+  return (await apiClient.get("datasets/")).data?.results;
 }
 
 export async function getDataset(datasetId: number): Promise<Dataset> {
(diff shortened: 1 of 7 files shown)


web/src/api/rest.ts
```diff
@@ -25,19 +25,19 @@ import type {
...
+  return (await apiClient.get("basemaps/")).data?.results;
 }
 
 export async function createBasemap(basemap: Basemap): Promise<Basemap> {
   return (await apiClient.post("basemaps/", basemap)).data;
 }
 
 export async function getProjects(): Promise<Project[]> {
-  return (await apiClient.get("projects/")).data.results;
+  return (await apiClient.get("projects/")).data?.results;
```
BryonLewis: Do you know why this was failing before?  If these endpoints expect a 'results' key on the data object I don't know if you want to use optional chaining to suppress the error.  That may hide the error in the future instead of surfacing some real problem.  This also means that all type definitions for these functions with the optional chaining are technically wrong because the Promise can return `undefined` thus making any error with the request/server look like some other error.

If the error is related to this sentry issue: https://kitware-data.sentry.io/issues/7371267566/?project=4511108704501760&query=&referrer=issue-stream
It looks like the actual problem is a http 401 response code and not the results being empty.  So maybe a check to make sure the response is a http 200 before accessing data or probaly better to see why the client is making a call to basemaps with (what looks like) an unauthenticated user.

Update:  I just pulled and tested master, it loooks like it is called before login so that is why I'm assuming the error is happening.

web/src/components/ControlsBar.vue
```diff
@@ -153,7 +153,17 @@ function createNewBasemapPreview() {
...
+      try {
+        const response = await fetch(newBasemapStyleJSON.value);
+        const content = await response.json();
+        jsonErrors.value = validateStyleMin(content);
+      } catch {
+        jsonErrors.value = [{ message: "Invalid URL." }];
+      }
+    } else {
+      jsonErrors.value = validateStyleMin(newBasemapStyleJSON.value);
+    }
```
BryonLewis, line 167: More a question, should these errors be surfaced to the user if the basemap fails to load?
Side note: thinking about this now should this be debounced or only checked on submission (enter) instead of watching the json string and on each character it sends a request?  Relatively minor thing but maybe something to think about if you're touching this section of code.

BryonLewis (changes requested):
The main thing is that the sentry error, I think removing the attempt to access the endpoint before logging in would be the proper fix instead of the optional chaining on the request results.

The other comment is more a question and can be ignored mostly because it's a niche feature I don't think mainly people would be using too often.

web/src/api/rest.ts
annehaley: You're right, the optional chaining was a lazy fix. I wasn't able to replicate the error so I tried to treat the symptom rather than the cause. But since you pointed out the fact that the basemap fetch can happen before authentication, that gave me a better idea.

 In [9e73157](https://github.com/OpenGeoscience/geodatalytics/pull/386/commits/9e7315720178a86e85b02a62e83de87613e7f1cb) I undid the addition of the optional chaining operators and tried something else. It looks like `fetchAvailableBasemaps` is called when the map component is mounted (it's used in the `handleMapReady` function passed to the `map-ready-a` and `map-ready-b` handlers on `ToggleMap`, which is [called when the component is mounted](https://github.com/BryonLewis/vue-maplibre-compare/blob/6587b337d4600b180a817b598cb5a6530c5111fb/src/components/ToggleCompare.vue#L547). I'm thinking a better solution is to conditionally render the map only when authentication is complete by using `v-if="appStore.currentUser"`. Lmk your thoughts on this fix.

web/src/components/ControlsBar.vue
annehaley: Good point, I added a debounce in [589031d](https://github.com/OpenGeoscience/geodatalytics/pull/386/commits/589031d4729c1b71b8b09e9cafb5baad46890f8f)

web/src/api/rest.ts
BryonLewis: Yeah this sounds good.  If you want the login page to have a default background map you could create another component with just the standard basemap and use that as as v-else before the user is logged in.

BryonLewis (approved):
works now, if you really want a background map before logging in, you can add it here before merging

annehaley:
Thanks, I added a default map (with a dark and a light mode) that shows when not authenticated.

BryonLewis (approved):
👍

PR #390, opened 2026-05-22
Use the standard prefix `VITE_` for Vite env vars

brianhelba:
`VUE_APP_` was used in Vue CLI, but [Vite uses only the prefix `VITE_`](https://vite.dev/guide/env-and-mode#env-variables). The extra `APP_` is pointless.


brianhelba:
Note, the Terraform changes need to be applied first (the failed build is an intended consequence of the new build check in this PR).

I can update CF Pages manually to get the build successfully passing, and to ensure there are no race conditions in deployment.

annehaley (approved):

PR #394, opened 2026-05-27
Enable ESLint rule `vue/return-in-computed-property`

brianhelba:
https://eslint.vuejs.org/rules/return-in-computed-property


brianhelba:
Actually depends on #395 to be resolved first.

brianhelba:
Ready for review.

I verified that the places these computed properties were used are safe for them to be `undefined` (though these were always actually returning `undefined`, we're just being more explicit about it now).

annehaley (approved):

PR #395, opened 2026-05-27
Enable ESLint rule `vue/no-async-in-computed-properties`

brianhelba:
https://eslint.vuejs.org/rules/no-async-in-computed-properties

### DIFF
--- a/web/src/components/sidebars/LayerStyle.vue
+++ b/web/src/components/sidebars/LayerStyle.vue
@@ -110,19 +110,23 @@ const showVectorOptions = computed(() => {
   return frames.value.some((frame) => frame.vector);
 });
 
+watch(
+  () => currentFrame.value?.vector,
+  async (vector) => {
+    if (vector && !vector.summary) {
+      vector.summary = await getVectorSummary(vector.id);
+    }
+  },
+  { immediate: true },
+);
+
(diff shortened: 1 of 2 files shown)


web/src/components/sidebars/LayerStyle.vue
```diff
@@ -110,19 +110,23 @@ const showVectorOptions = computed(() => {
   return frames.value.some((frame) => frame.vector);
 });
 
+watch(
+  () => currentFrame.value?.vector,
+  async (vector) => {
+    if (vector && !vector.summary) {
+      vector.summary = await getVectorSummary(vector.id);
```
brianhelba, line 117: Setting this will then cleanly force re-computation of `vectorProperties`.

web/src/components/sidebars/LayerStyle.vue
```diff
@@ -110,19 +110,23 @@ const showVectorOptions = computed(() => {
...
+    }
+  },
+  { immediate: true },
```
brianhelba, line 120: This will run unconditionally, to ensure it's bootstrapped. Previously, the first access of `vectorProperties` was triggering it to run.

annehaley (approved):

PR #396, opened 2026-05-27
Enable ESLint rule `vue/no-use-v-if-with-v-for`

brianhelba:
https://eslint.vuejs.org/rules/no-use-v-if-with-v-for

The change here satisfies the rule, but also removes a somewhat awkward pattern where a `v-for` was being abused to set a temporary variable. The core issue the lack of a `v-let` syntax for setting scoped template variables (which is [a requested feature](https://github.com/vuejs/rfcs/discussions/505)). However, in removing the `v-for` abuse, we are forced to call `findVectorProperty(filter)` several times in the rendered template, which looks bad, but should be quite cheap (particularly in comparison to the amount of Javascript that's ultimately run in the Vue engine to render any parts of a template).

### DIFF
--- a/web/eslint.config.ts
+++ b/web/eslint.config.ts
@@ -23,7 +23,6 @@ export default defineConfigWithVueTs(
       // Temporary ignores until rules can be fixed
       "vue/require-v-for-key": "off",
       "vue/valid-v-for": "off",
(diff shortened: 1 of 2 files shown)


annehaley (approved):
I was able to test the filter features locally and confirmed the behavior is unchanged.

PR #397, opened 2026-06-05
Multiframe scrubbing improvements

### LINKED ISSUE #126: Layer Frame Preloading
For large layers like the flood simulations, we should create an option for the user to preload all of its frames in the background so that switching between frames is more seamless.

BryonLewis:
Resolves #126
## Summary

Improves multiframe raster scrubbing by fixing RGB detection for sequential rasters, adding server-generated frame preview images, and crossfading from preview to tiles on the client for smoother frame transitions.

### Sentinel downloader (`scripts/sentinelDownload/`)
- Supports creating multiframe single-image Sentinel raster datasets via `--single-file` (uses `gdal_translate` to append frames as subdatasets)
- Additional CLI arguments for date range, cloud cover, output naming, and clip size
- Generates an ingest JSON alongside clipped GeoTIFFs for direct import into Geoodatalytics

### Dataset / RGB fixes
- **Default `band:1` injection removed** when no `source_filter` is supplied — this was preventing Boston ortho imagery from being recognized as RGB
- **`band:1` is still applied when explicitly defined** in the style spec
- **Multiframe rasters no longer render as grayscale** — `frame` and `band` are extracted from the style JSON and passed as separate query parameters (outside `style`) so django-large-image can auto-detect RGB imagery

### `RasterFramePreview` model
- Foreign keys to `LayerStyle` and `LayerFrame` (unique per style/frame pair)
- Stores `width`, `height`, georeferenced `bounds`, and an S3 preview image
- Lifecycle `status`: `creating` | `regenerating` | `complete` | `failed`
- `style_fingerprint` (`sha256` of `repr_style_configs()`) guards against stale concurrent generation
- Cascades on frame or layer style deletion; S3/MinIO object deleted via `post_delete` signal

### `uvdat/core/frame_previews/`
- `raster_style.py` - Builds django-large-image–compatible style objects from `LayerStyle` DB records; keeps `frame`/`band` outside `style` 
- `preview_regeneration.py` - Marks previews stale, supersedes pending tasks, enqueues Celery regeneration on style save/ingest 
- `fingerprint.py` - Computes style fingerprint for concurrency control 
-  `types.py` - Shared `FramePreviewData` type 

### `uvdat/core/tasks/frame_preview.py`
- Celery task generates per-frame preview PNGs via django-large-image thumbnails
- Uses `raster_style.py` for styling; scales resolution to keep previews between ~1k–4k
- Writes `RasterFramePreview` rows with bounds and dimensions
- Updates `TaskResult` on completion (used for future client notification)

### Preview regeneration on style save
- `LayerStyle` create/update hooks call `invalidate_and_enqueue_previews()`
- checks fingerprint hash of the style to determine if new previews need to be regenerated
- Existing preview images are cleared immediately; rows marked `regenerating`
- Ingest command also triggers preview generation for multiframe raster styles
- Only `status=complete` previews with an image are serialized to the API

### API
- `/layers/` and `/layer-styles/` return `multiframe_previews` — ordered list of `{ bounds, width, height, url }` per raster frame (presigned S3 URLs)
- `preview_status` aggregated on layer and layer-style: `ready` | `generating` | `notready` (omitted when N/A)
- Queryset annotations avoid N+1 Python aggregation

### Tests
- `uvdat/core/tests/test_frame_preview.py` — model serialization, API fields, regeneration hooks, fingerprint guards, raster style helpers

### Frame Preview Logic
**Loading  Frame Preview Flow**
1. User changes frame -> `layerStore.updateLayersShown`
2. `styleStore.updateLayerStyles`
3. `framePreviewStore.showPreviewThenTiles`
   - Hide tile layer (`raster-opacity: 0`)
   - Show georeferenced preview image
   - Wait for tile source to finish loading
   - Crossfade preview → tiles once loaded
   - Preload adjacent frames

**`web/src/utils/framePreviewLayer.ts` Functions**
- `upsertPreviewLayer` — fetch preview URL, add/update MapLibre source + raster layer
- `hidePreviewLayer` — set visibility to `none` during transition
- `removePreviewLayer` / `removePreviewLayersExcept` — cleanup; keep adjacent preloaded frames
- `waitForRasterSourceLoaded` — poll until tile source is ready (with timeout)
- `fadeRasterOpacities` — animated crossfade between preview and tile layers
**`web/src/store/framePreview.ts` Functions**
- `prefetchLayerPreviews` — preload all preview images for a layer
- `showPreviewThenTiles` — main preview → tile transition orchestration
- `dismissPreviewForLayer` — skip previews while editing layer styles; restore tile opacity

**Style editing integration**
- `styleStore.setLayerStyleEditing` dismisses previews when entering style edit mode
- Previews re-enabled via `updateLayerStyles` after exiting edit mode
- `LayerStyle.vue` watches active layer to toggle editing state

**Creating/Editing and Saving Style Flow**
1. User finishes editing and clicks Create Style or Save in `LayerStyle.vue`
2. 1createLayerStyle1 (new) or 1updateLayerStyle1 (existing) -> POST/PATCH /api/v1/layer-styles/
3. Backend (`LayerStyleSerializer.create` / `.update`)
    1. inside those above functions it calls `invalidate_and_enqueue_previews` for multifram rasters only
        1. This function starts by computing `style_fingerprint` - hash for comparing to see if the style changed
        2. `previews_current_for_fingerprint` - if the fingerprint is the same (same style) don't invalidate and generate new previews, instead just return.
        3. Marks all the existing frame previews as regenerating/creating to indicate they are stale
        4. If there are any pending preview tasks it will mark them as superseceded to prevent updating the previews for this style
        5. Create a TaskResult for generating the new previews then execute the task
    2. return back to the client the new layerstyle with the `preview_status` of `not ready` to indicate that the previews are being created.
4. Client Side Response
    1. `markStyleSavedAndInvalidatePreviews(style)` - clear the existing previews for the style
    2. Apply saved style to `styleStore.selectedLayerStyles` with `multiframe_previews: undefined`
    3. `framePreviewStore.dismissPreviewForLayer`
        1. Bump transition generation (cancels in-flight crossfades)
        2. Remove all preview MapLibre layers
        3. Restore tile layer opacity, user sees tiles styled with the new spec
refreshLayer() -> layerStore.fetchAvailableLayer (refreshes layer metadata if default changed)
    4.  While the user still has the editor window open
        - showPreviewThenStyles is skipped so they see the style that is being edited and not previews
6. The generating previews task is async, when complete there is a websocket notification of `framePreview` that lets the client know that new previews are available
    1. `framePreviewStore.onPreviewTaskComplete(task)`
    2. `getLayerStyle(layer_style_id)` - refetch style with `preview_status: "ready"` + multiframe_previews
    3. Update selectedLayerStyles for every selected copy still using that style
    4. Mirror onto layer object if style is default
    5. `prefetchLayerPreviews` - warm image URLs
    7. If not in style edit mode: `showPreviewThenTiles` -> preview overlay + crossfade back to tiles


BryonLewis:
be able to extract the opacity and default frame when calculating the style fingerprint.

BryonLewis:
add in sorting to the dictionary for the styles to resolve any ordering of hashes.

uvdat/core/frame_previews/preview_regeneration.py
```diff
@@ -0,0 +1,152 @@
...
+
+def supersede_pending_preview_tasks(layer_style_id: int) -> None:
+    TaskResult.objects.filter(
+        task_type="frame_preview",
+        completed__isnull=True,
+        inputs__layer_style_id=layer_style_id,
+    ).update(
+        completed=timezone.now(),
+        status="Superseded by newer style save.",
+    )
```
annehaley, line 24: I see that this will cause a pending task to exit with a skip message right after it starts, but is there anything that can be done about ongoing tasks? Would it make sense to add checks throughout the task to see if the preview generation should be cancelled?

uvdat/core/frame_previews/fingerprint.py
```diff
@@ -0,0 +1,20 @@
...
+
+if TYPE_CHECKING:
+    from uvdat.core.models import LayerStyle
+
+
+def _fingerprint_payload(configs: dict) -> str:
+    normalized = dict(configs)
+    normalized["default_frame"] = int(normalized["default_frame"])
+    normalized["opacity"] = float(normalized["opacity"])
+    return json.dumps(normalized, sort_keys=True, default=str)
```
annehaley: You already noted this in your comments, but I'll make a review comment anyway to ensure these are addressed. We can ignore `default_frame` and `opacity` here because changes to these would not require preview regeneration. And we need to make sure that the order of the dict is enforced before comparing the hashes of the json.

uvdat/core/frame_previews/preview_regeneration.py
```diff
@@ -0,0 +1,152 @@
...
+    fingerprint = style_fingerprint(layer_style)
+    if previews_current_for_fingerprint(layer_style, fingerprint):
+        return None
+
+    mark_previews_regenerating(layer_style, fingerprint)
+    clear_style_preview_instance_cache(layer_style)
+    supersede_pending_preview_tasks(layer_style.id)
+
+    result = TaskResult.objects.create(
+        name=f"Frame previews: {layer_style.name}",
```
annehaley: Very minor, but it may be worthwhile to include the layer name in the result's name, too. The style name is only unique among styles on the same layer, so you could have the same style name on two different layers.

uvdat/core/frame_previews/preview_regeneration.py
```diff
@@ -0,0 +1,152 @@
...
+        task_type="frame_preview",
+        project=layer_style.project,
+        inputs={
+            "layer_style_id": layer_style.id,
+            "layer_id": layer_style.layer_id,
+            "fingerprint": fingerprint,
+        },
+    )
+
+    from uvdat.core.tasks.frame_preview import generate_layer_style_previews  # noqa: PLC0415
```
annehaley: Is this a lazy import due to circular import errors? A comment explaining why would be useful here.

uvdat/core/frame_previews/raster_style.py
```diff
@@ -0,0 +1,148 @@
...
+RASTER_SOURCE_FILTER_KEYS = frozenset({"frame", "band"})
+
+
+def _hex_to_rgb(color: str) -> tuple[int, int, int]:
+    color = color.lstrip("#")
+    return int(color[0:2], 16), int(color[2:4], 16), int(color[4:6], 16)
+
+
+def _rgb_to_hex(rgb: tuple[int, int, int]) -> str:
+    return "#{:02X}{:02X}{:02X}".format(*rgb)
```
annehaley: I see that this file is copying logic that previously only existed client-side. It makes sense that we need it on the server now too. But can we avoid having the same logic duplicated in two places (with two different languages)? That makes it harder to maintain. Maybe we could add an API endpoint on the style viewset that can provide the query string for a given style and use that result on the client side?

uvdat/core/management/commands/ingest.py
```diff
@@ -385,10 +419,16 @@ def ingest(*, file_path, replace, clear, skip_cache):
...
+        datasets, file_path, replace=replace, skip_cache=skip_cache
+    )
     click.echo("Ingesting Projects:")
     ingest_projects(projects, replace=replace)
     click.echo("Ingesting Charts:")
     ingest_charts(charts, replace=replace, skip_cache=skip_cache)
 
+    preview_count = generate_ingest_multiframe_previews(converted_dataset_names)
+    if preview_count:
+        click.echo(f"Generated multiframe raster previews for {preview_count} layer style(s).")
```
annehaley, line 426: Rather than doing this as a step of the ingest management command, this should be a part of the dataset conversion task so that it will happen for datasets uploaded via the UI or created as a result of an Analytics task. I would recommend calling this at the end of `create_layers_and_frames` in `uvdat/core/tasks/dataset.py`.

uvdat/core/models/layer.py
```diff
@@ -25,6 +30,41 @@ class Layer(models.Model):
     def __str__(self):
         return f"{self.name} ({self.id})"
 
+    def multiframe_raster_frames(self):
+        prefetched = getattr(self, "raster_frames", None)
+        if prefetched is not None:
+            return prefetched
+        return list(
+            self.frames.filter(raster__isnull=False).select_related("raster").order_by("index")
+        )
```
annehaley, line 39: This implementation doesn't appear to be limited to multiframe layers. This could return a single raster frame, so let's just rename this without "multiframe".

uvdat/core/models/styles.py
```diff
@@ -224,6 +231,32 @@ def repr_style_configs(self):
...
         }
 
+    @staticmethod
+    def serialize_frame_preview(preview: RasterFramePreview) -> FramePreviewData:
+        return FramePreviewData(
+            url=preview.image.url,
+            width=preview.width,
+            height=preview.height,
+            bounds=preview.bounds,
+        )
```
annehaley, line 244: Why not use a DRF serializer for this?

uvdat/core/models/task_result.py
```diff
@@ -16,6 +19,24 @@
 from .project import Project
 from .querysets import ProjectQuerySet
 
+logger = logging.getLogger(__name__)
+
+# Set while running in a context (e.g. `manage.py ingest`) where TaskResult
+# WebSocket notifications are meaningless: tasks run synchronously with no
+# client session listening. When set, result_post_save skips the push entirely
+# rather than attempting it and logging a warning.
+_suppress_notifications: ContextVar[bool] = ContextVar("suppress_task_notifications", default=False)
```
annehaley, line 28: This is an interesting addition that I had never considered. I do think it's useful, but I think we should be applying this context differently. Right now, I see that there's one place where you've applied this context, in `ingest.py`:
```
with suppress_task_notifications():
    invalidate_and_enqueue_previews(style, asynchronous=False)
```
If we do end up moving the preview generation to the dataset conversion task (see earlier comment), we can't always apply this context. But we can apply it if `asynchronous=False`. My suggestion is that we add the `with suppress_task_notifications():` line inside `invalidate_and_enqueue_previews` right before we call `.apply()`. 

In fact, we can do that in all three places where we run a task synchronously. If you search the code for `.apply(`, you can add the context wrapper around all three places, which includes the `convert_chart` and `convert_dataset` tasks.  This should make the suppression behavior more consistent for all synchronous tasks.

uvdat/core/rest/dataset.py
```diff
@@ -43,7 +44,9 @@ def tags(self, request, **kwargs):
     @action(detail=True, methods=["get"])
     def layers(self, request, **kwargs):
         dataset: Dataset = self.get_object()
-        layers = list(dataset.layers.all())
+        for layer in Layer.objects.filter(dataset=dataset, default_style__isnull=True):
+            layer.ensure_default_style()
```
annehaley, line 48: I'm concerned about having a db-write side effect within a get endpoint. Is there anywhere else we could put this statement to ensure that layers have a default style? Same goes for the `ensure_default_style` call in `uvdat/core/rest/layer.py`. 

Your comment about `ensure_default_style` explains that a layer can be left without a default style if its existing default style is deleted, so maybe we can just put that function call in a delete signal?

uvdat/core/tasks/frame_preview.py
```diff
@@ -0,0 +1,419 @@
...
+    image = Image.open(io.BytesIO(png_bytes))
+    return png_bytes, image.width, image.height, _preview_bounds(source)
+
+
+def ensure_default_layer_style(layer: Layer, project: Project) -> LayerStyle:
+    """Return the project's Default style for a layer, creating it on ingest if needed."""
+    style = LayerStyle.objects.filter(layer=layer, project=project, name="Default").first()
+    if style is None:
+        style = LayerStyle.objects.create(name="Default", layer=layer, project=project)
+        style.save_style_configs(DEFAULT_MULTIFRAME_RASTER_STYLE_SPEC)
```
annehaley, line 187: I'm not sure I fully understand why we need to create default style objects. The current behavior allows for the absence of any created styles on a layer; in that case, raster layers will just use the large-image defaults. Is there a reason we can no longer use that pattern? If it's only for the foreign key relationship on the preview, we could just allow that field to be null.

uvdat/core/tasks/frame_preview.py
```diff
@@ -0,0 +1,419 @@
...
+    """
+    return style_fingerprint(layer_style) == fingerprint
+
+
+def _open_task_result(result_id: int | None, layer_style_id: int) -> TaskResult | None:
+    """Load the TaskResult for this run, or None if it was already superseded."""
+    if result_id is None:
+        return None
+
+    result = TaskResult.objects.filter(id=result_id).first()
```
annehaley, line 213: Just use `TaskResult.objects.get(id=result_id)`

web/src/store/framePreview.ts
```diff
@@ -0,0 +1,469 @@
...
+
+    removePreviewLayer(map, layerKeyValue, frameIndex);
+    activePreviewByLayerKey.delete(layerKeyValue);
+    clearPreviewDisplayed(layerKeyValue);
+  }
+
+  async function showPreviewThenTiles(layer: Layer) {
+    const layerStore = useLayerStore();
+    const mapStore = useMapStore();
+    const styleStore = useStyleStore();
```
annehaley, line 266: For these other stores that you use in this store's functions, just get them all at the top of the store definition. That's what we do in all the other stores.

annehaley (changes requested):
Since this is a large PR, I'll do my reviews in two rounds. This first round is just from reading the diff. After these comments are addressed I'll do a second round where I try it out locally and can comment on the behavior.

uvdat/core/frame_previews/fingerprint.py
BryonLewis: Popped the `default_frame` and `opacity` keys from the style JSON specification in https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/ea355071f212dfc377fa8e0f4f7c015f7991a39a

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: Added in `{layer_style.layer.name} -  {layer_style/name}` to the TaskResult for more information in: https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/3bc788f8b9c5494f8578c7598f380764b9c5c7cf

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: That I think was a result of an earlier circular import issue during testing.  It shouldn't be an issue now so I'm moving to the top of the file in https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/46626be1510626bd7a595115f96e09af547234e8

uvdat/core/tasks/frame_preview.py
BryonLewis: I think it keeps some of the logic simpler if we have an explicit default layer style after data is ingested.

The preview system is essentially keyed on the `LayerStyle`.  We can change it but we would need to add logic branches to handle fingerprinting, task invalidation, the `preview_status` in the HTTP response.  I.E. conditionals of `if layer.default_style is None:` then do a slightly different thing using `DEFAULT_MULTIFRAME_RASTER_STYLE_SPEC`

uvdat/core/tasks/frame_preview.py
BryonLewis: I'm using the filter there to prevent a `DoesNotExist` error that would happen with `get` where instead I want `None`.  When the result is None it means that the task result has been superseded by another task. I.E while generating previews a user chained their mind so it started another task for generating previews.

I could update it to use try block with catching `DoesNotExist`.  I could also add a comment as to why I'm using filter here.

uvdat/core/models/styles.py
BryonLewis: I kept the TypedDict helper because this is a small nested read object (four fields) with no write/validation path, and FramePreviewData is easier to reason about as a type hint at the model call sites than DRF serializer.

uvdat/core/frame_previews/raster_style.py
BryonLewis: What we have now:

1. The client side has a query builder for large-image rasters so it can take colormaps/filters and create a preview while editing
2. The internal storage of these colormaps and filters is stored in a separate data structure which is passed to the server when you save a style
3. This JSON style specification is broken into pieces and saved as ColormapConfig, SizeConfig, FilterConfig and others

So when we request a style from the server for the client it needs to take the style JSON and build back up the query params for raster large-images.

Should during saving should we also save the large-image query params that the client has already created.  The previews then can use that and break it up into it's params for large-image instead of trying to re-interpret the ColormapConfig and FilterConfig and generate the query params?

annehaley: Yeah, that idea is cleaner. We can add a `tile_query` field to the `LayerStyle` model (nullable, since vector styles won't need it). Then we don't need any new logic on the server side to convert the db representation of the style into a query.

uvdat/core/models/styles.py
annehaley: Ok, that makes sense. I see now that we never need to write this object.

uvdat/core/tasks/frame_preview.py
annehaley: Ok that makes sense. A comment would be sufficient, then.

uvdat/core/tasks/frame_preview.py
annehaley: With your current approach, we need to ensure that every layer has a default style created. We only need them to exist for raster layers, but for the sake of consistency we make them for every layer. You've added this step in the ingest process, and you also ensure that step is completed for already-existing layers. You currently do this in the endpoint where we get all layers for a target dataset, and I've noted in my previous comment that I don't think we should do that in a get request. If you can find another place to do that, I think your approach is acceptable, but design-wise I still think a None style makes more sense. 

How much more complex would it be to support a None style in those places you mentioned, compared to creating these default style objects? They seem equally complex to me.

BryonLewis: Yeah I completely agree with moving the creation out of the GET endpoint comments (that was an oversight on my part).

I think it would require that within the layer need some default Query URL (like in the other comment) for default_style and possible a fingerprint style spec for checking to make sure we really need to create new empty previews on save.
- Then in the code it would need changes, mostly more arguments because I access the layer a lot through the layer_style.layer:
    - `invalidate_and_enqueue_previews` - the layer_style passed in could be null meaning that the fingerprint would be some default fingerprint.  So we need to pass in the Layer as well to this function because before we would access the layer by using the connected layer_style.layer
    - When `mark_previews_regenerating` it would have the new layer and the layer_style (which could be null) and create RasterFramePreview objects where the layer_style foreign key is null.
    - `clear_style_preview_instance_cache` would also need the Layer and the layer_style passed in the function to remove the previous cached prefetched results
    - `supersede_pending_preview_tasks` would need the layer id in addition to the layer_style_id and the inputs for the task for the TaskResult would require the layer_id as well as the layer_style_id to be added to invalidate any previous tasks creating a style for the default style
    - when passing into `generate_layer_style_previews` (the task for creating the previews) the layer_id would have to be added along with the layer_style id.
- Inside of the preview generation task:
    - the `_PreviewGenerationContext` would now need the layer_id as well as the layer_style_id
    - in the creation of a frame preview `_process_frame_preview` it would now use a git based on the layer_id, the layer_style (null being valid) and the frame for replacing/creating a new Preview object
- Access for REST Responses:
    - `preview_status_for_style(layer_style)` - this was used before to determine if previews were ready for a style.  I may need a whole seprate function of `preview_status_for_empty_style(layer_id)` that would check the RasterFramePreviews that match the layer_id and see if they are all ready for preview.  In the original function layer_styles now had a foreign key to layer_style so I could use `layer_style.frame_previews` and see if they were all ready.  Now I think I need to update this function to use the style_previews (foreign key into layerFrame).  So this seprate function would take a layer -> get the layer_frames for it and then filter RasterFramePreviews for any matching LayerFrame Ids that have a null layer_style set.
    - The Layer serializer itself would return None if the obj.defaul_style_id is None.  So instead we need to get the layer_frames from the layer, then filter to see what RasterFramePreviews have a null layer style and match the layer_frame for the null case of default style and return them as the previews.  Again we can't utilize the foreign key of `layer_style.frame_previews`
    - There may be some abstraction that can be done in the two above functions to create a single way to take a layer with no default style and return the frame previews to check their status or return their presigned URLS for serialization I just thought about it after walking through the logic.
- One thing I didn't consider is that I assume the empty default_layer_style for the raster frame previews just sticks around and is never deleted until the layer itself is deleted.  That way if the user ever deletes the styles it would have this set of styles to fall back on.
    
That's a rough 15-20 minute analysis of what may need to change with a None default_style.  Most of it has to do relying on the foreign key relationships between LayerStyle and RasterFramePreview  (layer_style.frame_previews) being broken as well as utillizing layer_style.layer for reverse lookup of the layer.

uvdat/core/rest/dataset.py
BryonLewis: This is an oversight on my part.  I need to do a bit more thinking and it relates to the other questiona bout having a default_style LayerStyle object for use with the previews.
I haven't made a decision but I'm weighing the options
- Move the creation of the frame-previews and the default style for multiframe raster into the dataset ingestion task script
    - additional write a migration that will create a layerstyle for any multiframe rasters that don't have one already
- Accept the case of not having a default_style and having the logic to handle this and properly generate one if none exists (probably again in a migration).

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: There is somewhat of a check during the task creation process.  It goes through a loop of the frames and before it does each frame it calls `_process_frame_preview` which will return with an outcome variable. The outcome variable could become "superseded" if the current LayerStyle's fingerprint doesn't match the task's stored fingerprint.  This is an indication that the task is 'dirty' meaning another frame preview task has started so it will exit immediately before writing the image.

uvdat/core/models/task_result.py
BryonLewis: Added the suppression to other locations that use `.apply` and have a sync and async path.  I'll be moving the `invalidate_and_enqueue_previews` out of the ingest.py and into the dataset conversion process in a future PR.
This is the pr that has the base changes: https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/c4525bdb48fda3b9f79e98505a31796a820bc86c

uvdat/core/tasks/frame_preview.py
BryonLewis: added a descriptivie comment in: https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/515223ea1146b9fb9391c4e0a4e6dab00fc15bb5

uvdat/core/frame_previews/raster_style.py
BryonLewis: I've gone ahead and done that by adding a `raster_style_params` to the layerStyle model and redid the migration: [261c649](https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/261c6496ee326d7d99cfb6dc88f8a9162b92123b)

Removed the previous functions for computing the raster style and added functionality to save raster_style_param when updating a style as well as have the taskf or preview generation use the `raster_style_params`: https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/28ff79c20247353acb4bcf32590a63ee1ccd493d

Update the client to send the raw raster style query instead of the StyleSpec when saving/updating a style: https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/9c589d8634466bd766b0ac1445168f7155b90cd2

Updating tests: https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/a85378813ab615dd4d74136d53e138417033fedc

The flood simulation creates it's own stylespec on completion.  Added the `raster_style_params` to it so it can properly generate previews without recomputing the query from the stylespec: https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/2ee4537f45b984f0117813857fc23fc0b1f1f23a

Updated the fingerprinting logic to use the `raster_style_params` instead of the older converted `style_spec`: https://github.com/OpenGeoscience/geodatalytics/pull/397/commits/fa6e88d14ea602d0d1a830c77a534cc2465b17c7

BryonLewis:
So two of the bigger comments were about generating previews at dataset conversion time, and about requiring a default `LayerStyle` just to hold those previews. Those issues are linked to how `RasterFramePreview` relates to `LayerStyle` today.

`RasterFramePreview` currently has foreign keys to both `LayerStyle` and `LayerFrame`. That gives a direct ownership path: create/update a style -> regenerate its previews; delete a style -> cascade the preview rows (and their images).
`LayerStyle` is also tied to a `Project`. That makes it hard to create default raster frame previews during dataset conversion, because the dataset is not linked to a project until after conversion completes.

A style-owned preview model is natural when a layer in a project can have multiple `LayerStyle`s, each with its own appearance and preview set. The downside is that layers without a default `LayerStyle` cannot expose previews, so ingest ends up creating a Default style mainly as a place to hang them.
There are a few ways forward:

### Option 1 — Flat / content-addressed previews
Key `RasterFramePreview` by `(LayerFrame, style_fingerprint)`, where the fingerprint is a hash of the `raster_style_params` used to render the PNG. Optionally store the params JSON on the row for debugging/re-render.
`LayerStyle` no longer owns preview rows. Looking up previews means: hash the style’s `raster_style_params` -> fetch complete previews for that layer’s frames. Ingest can write the empty-params (`{}`) preview set at conversion time with no `Project` or `LayerStyle`.
**Pros**
- Default empty-style previews can be created at dataset conversion
- A dataset/layer can show default previews without any `LayerStyle`
- Styles that share the same params on the same frames share one preview set (no duplicate PNGs)
**Cons**
- Lookup is by fingerprint (+ frame), not a direct `LayerStyle` FK (still cheap with a unique index, but different from today)
- Deletion/GC is no longer a simple cascade: removing a style must not delete cache rows still used by another style with the same params (refcount, lazy GC, or accept orphans).  Really the deletion of a layer is what would delete the preview styles
- Access control must still be enforced via dataset/layer (or style) permissions — the fingerprint is a cache key, not an authorization mechanism. Sharing previews across projects that already share a dataset is expected and fine if those checks stay correct

### Option 2 — Dataset-scoped default `LayerStyle` (`project=NULL`)
Keep preview ownership on `LayerStyle`, but allow one special default style per layer with `project=NULL` (dataset-scoped). Ingest/conversion creates that style and its previews once. Any project that imports the dataset sees that default immediately.
Editing the shared default should **not** mutate it for every project. Project users get copy-on-write: modifying the default forks a new project-scoped `LayerStyle`, and further preview regen hangs off that fork. The null-project default stays read-only for normal project members (dataset owners/admins may still update it deliberately).
**Pros**
- Preserves direct FKs, easy lookup, and CASCADE cleanup for user/project styles
- Deleting a layer/dataset still cleans up the default style and its previews
- Enables default previews at conversion time without inventing a separate cache model
- Avoids duplicating the *default* style/preview set across projects that share a dataset
**Cons**
- Introduces a special case: `LayerStyle` is usually project-scoped, but the default is not
- Still forces creating a `LayerStyle` at dataset/layer creation for multiframe rasters
- Queryset/permission rules must explicitly include null-project defaults for projects that have the dataset
- User forks can still create duplicate styles/previews across projects when people customize away from the default
- Preview tasks/`TaskResult` currently assume a project; null-project defaults need a clear rule for task scoping/notifications


After writing this out and reviewing, I'm going to create a PR into this branch that implements option 1 utilizing the fingerprint instead of having a foreign key reference to see how that works and if it is a better mechanism.

BryonLewis:
closing in favor of #427

PR #399, opened 2026-06-05
Enable ESLint rules `vue/valid-v-for` and `vue/require-v-for-key`

brianhelba:


web/eslint.config.ts
```diff
@@ -20,9 +20,6 @@ export default defineConfigWithVueTs(
       "vue/valid-v-slot": ["error", { allowModifiers: true }],
       // `any` is used everywhere and will be difficult to eliminate
       "@typescript-eslint/no-explicit-any": "off",
-      // Temporary ignores until rules can be fixed
-      "vue/require-v-for-key": "off",
-      "vue/valid-v-for": "off",
```
annehaley, line 25: Great, looks like these were the last ones that needed to be addressed. Tysm for doing all that cleanup!

annehaley (approved):

PR #400, opened 2026-06-05
Set Redis max memory policy

### LINKED ISSUE #398: Set Redis Max Memory Policy
Since we started using Redis for our large-image cache, we often encounter container crashes due to exceeding the default max memory. We need to change the max memory policy in both dev (`docker-compose.yml`) and prod environments (`terraform/django.tf`).

```
["redis-server", "--maxmemory", "4096MB", "--maxmemory-policy", "allkeys-lru"]
```

annehaley:
Resolves #398 

This PR changes the command for the Redis container in development to the following:
```
["redis-server", "--maxmemory", "4096MB", "--maxmemory-policy", "allkeys-lru"]
```

For production, we can't set the max memory size of the Heroku Redis add-on without upgrading the plan (we currently use the mini plan), but we can set the max memory policy to `allkeys-lru`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,6 +48,7 @@ services:
 
   redis:
     image: redis:alpine
+    command: ["redis-server", "--maxmemory", "4096MB", "--maxmemory-policy", "allkeys-lru"]
     healthcheck:
       test: ["CMD", "redis-cli", "ping"]
       start_period: 30s
--- a/terraform/django.tf
+++ b/terraform/django.tf
@@ -40,6 +40,10 @@ module "django" {
 resource "heroku_addon" "redis" {
   app_id = module.django.heroku_app_id
   plan   = "heroku-redis:mini"
+
+  config = {
+    maxmemory_policy = "allkeys-lru"
+  }
 }
 
 output "dns_nameservers" {


docker-compose.yml
```diff
@@ -48,6 +48,7 @@ services:
 
   redis:
     image: redis:alpine
+    command: ["redis-server", "--maxmemory", "4096MB", "--maxmemory-policy", "allkeys-lru"]
```
brianhelba, line 51: Ideally we'd set this in CI too, but GH Actions doesn't let you pass a custom `command` to a service.

annehaley: Yeah, that's fine. It wouldn't make a difference for our tests anyway.

BryonLewis (approved):

brianhelba (changes requested):
I have concerns about how this will affect the use of Django Channels, which uses Redis to mediate websocket connections.

I think a better policy may be `volatile-lru`, but that might require cooperation from `large_image`. I'm happy to help investigate more.

manthey:
> I have concerns about how this will affect the use of Django Channels, which uses Redis to mediate websocket connections.
> 
> I think a better policy may be `volatile-lru`, but that might require cooperation from `large_image`. I'm happy to help investigate more.

`volatile-lru` requires that entries have a TTL.  large_image uses redis (or other tile caches) strictly as LRU -- there is no reason to evict tiles except for the memory area being full and the then LRU us a sensible policy.  If websocket connections are more recently used than the speed at which tiles are cached, this won't functionally be a problem, but fundamentally, these two sets of data have very different concerns.  The correct architecture is really have two separate caches, one that is LRU for tiles and one that is ALWAYS as large as total possible websocket connections .

brianhelba:
#363 set our cache implementation to use [`large_image...RedisCache`](https://github.com/girder/large_image/blob/b8a0a2522325bcb76265d60999d5fe2f49dacdde/large_image/cache_util/rediscache.py#L31), which doesn't support setting TTL on keys.

However, we could:
* Switch back to using [`django_large_image...DjangoCache`](https://github.com/girder/django-large-image/blob/8f6c6aa8247e6161eaff99f93927cfd9259acbc4/django_large_image/cache.py#L10)
  * Just set `LARGE_IMAGE_CACHE_BACKEND = "django"` or unset it entirely as it's the `django_large_image` default
* `DjangoCache` will rely on Django's general cache framework
* Configure Django's cache framework [to use Redis](https://docs.djangoproject.com/en/6.0/topics/cache/#redis)
  * [Set a `TIMEOUT`](https://docs.djangoproject.com/en/6.0/topics/cache/#cache-arguments) to add by default to all keys
  * Use a more specific cache name, rather than making a `"default"` Django cache, in case other parts of Django want to use the `"default"` cache
    * You should be able to [set `LARGE_IMAGE_CACHE_NAME` accordingly](https://github.com/girder/django-large-image#optional-cache-configuration)

brianhelba:
I think `volatile-lru` gives us the balance we want. We can set the TTL value on the tile cache to be arbitrarily long, so tiles will practically only get evicted by memory pressure, not time. In this case, the TTL is still essential as a marker that the tile data doesn't need to be preserved for the application to function correctly.

manthey:
> I think `volatile-lru` gives us the balance we want. We can set the TTL value on the tile cache to be arbitrarily long, so tiles will practically only get evicted by memory pressure, not time. In this case, the TTL is still essential as a marker that the tile data doesn't need to be preserved for the application to function correctly.

I can't tell from how the redis docs are written if the eviction is anything that is past its TTL or anything that has a TTL at all.  If the former, then the shortest possible TTL would permit tile to be evicted (but redis would normally keep them since it takes effort to evict things).  If the latter, the a long TTL would prevent needless eviction.

manthey:
Redis says: "Use volatile-ttl if your code can estimate which keys are good candidates for eviction and assign short TTLs to them. Note also that if you make good use of key expiration, then you are less likely to run into the cache memory limit because keys will often expire before they need to be evicted.", so the large-image tiles should have very short TTLs.  See https://redis.io/docs/latest/develop/reference/eviction/

annehaley:
Closing this PR in favor of the suggested approach. #404 will revert back to using Django for the large_image cache, and we will later make an upstream change to large_image (tracked with [this issue](https://github.com/girder/large_image/issues/2081)) that will allow using Redis again with TTLs enabled

brianhelba:
Sounds good. I actually think that the large_image tiles could have an arbitrarily long TTL (like 1 year), as the mere presence of any TTL value will cause them to evictable under memory pressure.

PR #401, opened 2026-06-05
Switch `eslint-plugin-vue` rules to `flat/recommended`

brianhelba:
This is [used by BatAI](https://github.com/Kitware/batai/blob/bda6a6a535a478da9284addef827eeccceeb6462/client/eslint.config.ts#L17) and provides stronger error detection.

### DIFF
--- a/web/eslint.config.ts
+++ b/web/eslint.config.ts
@@ -11,16 +11,23 @@ export default defineConfigWithVueTs(
     name: "app/files-to-lint",
     files: ["**/*.{vue,ts,mts,tsx}"],
   },
+
   globalIgnores(["**/dist/**", "**/dist-ssr/**", "**/coverage/**"]),
-  ...pluginVue.configs["flat/essential"],
+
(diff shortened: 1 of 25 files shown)


web/eslint.config.ts
```diff
@@ -11,16 +11,23 @@ export default defineConfigWithVueTs(
...
+
   {
     rules: {
       // allowModifiers: Vuetify uses dot notation for column slots (v-slot:item.columnName)
       "vue/valid-v-slot": ["error", { allowModifiers: true }],
       // `any` is used everywhere and will be difficult to eliminate
       "@typescript-eslint/no-explicit-any": "off",
+      // Temporary ignores until rules can be fixed
+      "vue/no-template-shadow": "off",
+      "vue/require-default-prop": "off",
```
brianhelba, line 28: This shouldn't be too hard to fix by hand in a follow-up PR.

annehaley (approved):

PR #406, opened 2026-06-18
Ask Qwen Task

annehaley:
This PR adds a new analytics task called "Imagery: Ask Qwen". The task accepts three inputs: 1) an imagery raster, 2) a text prompt, and 3) a `max_tokens` value (1k - 10k). The task leverages our first Huggingface inference endpoint, sending a lower resolution version of the imagery to the endpoint along with the question and returning the response to the user.

The screenshots below show some example runs on the boston orthoimagery layer:
geodatalytics_imagery_ask_qwen_task
geodatalytics_imagery_ask_qwen_2

To support the large markdown responses, this PR makes a slight change to the Analytics Panel in the client, using [`vue-markdown-render`](https://www.npmjs.com/package/vue-markdown-render) for a new output type "markdown".

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,7 @@ tasks = [
   "pyvips==3.1.1.8.18.1",
   "uvdat-flood-sim[large-image-writer]==1.0.4",
   "xdg-base-dirs==6.0.2",
+  "huggingface-hub>=1.14.0",
 ]
 
 [dependency-groups]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ tasks = [
   "pyvips==3.1.1.8.18.1",
   "uvdat-flood-sim[large-image-writer]==1.0.4",
   "xdg-base-dirs==6.0.2",
-  "huggingface-hub>=1.14.0",
+  "huggingface-hub==1.14.0",
 ]
 
 [dependency-groups]
--- a/terraform/django.tf
+++ b/terraform/django.tf
@@ -20,9 +20,11 @@ module "django" {
   ec2_worker_ssh_public_key    = file("${path.module}/ssh-key.pub")
 
   additional_django_vars = {
-    DJANGO_UVDAT_WEB_URL          = "https://www.geodatalytics.kitware.com/"
-    DJANGO_DATABASE_POOL_MAX_SIZE = "12"
-    DJANGO_SENTRY_DSN             = "https://5302701c88f1fa6ec056e0c269071191@o267860.ingest.us.sentry.io/4510620385804288"
+    DJANGO_UVDAT_WEB_URL           = "https://www.geodatalytics.kitware.com/"
+    DJANGO_DATABASE_POOL_MAX_SIZE  = "12"
+    DJANGO_SENTRY_DSN              = "https://5302701c88f1fa6ec056e0c269071191@o267860.ingest.us.sentry.io/4510620385804288"
+    DJANGO_UVDAT_HF_NAMESPACE      = "Kitware"
+    DJANGO_UVDAT_HF_ENDPOINT_NAMES = "qwen=qwen3-5-9b-gguf-ulh,"
   }
   additional_sensitive_django_vars = {
     DJANGO_UVDAT_HF_TOKEN = var.DJANGO_UVDAT_HF_TOKEN
--- a/terraform/django.tf
+++ b/terraform/django.tf
@@ -24,6 +24,9 @@ module "django" {
     DJANGO_DATABASE_POOL_MAX_SIZE = "12"
     DJANGO_SENTRY_DSN             = "https://5302701c88f1fa6ec056e0c269071191@o267860.ingest.us.sentry.io/4510620385804288"
   }
+  additional_sensitive_django_vars = {
+    DJANGO_UVDAT_HF_TOKEN = var.DJANGO_UVDAT_HF_TOKEN
+  }
   django_cors_allowed_origins = [
     # Can't make this use "aws_route53_record.www.fqdn" because of a circular dependency
     "https://www.geodatalytics.kitware.com",
--- a/uvdat/core/tasks/analytics/imagery_ask_qwen.py
+++ b/uvdat/core/tasks/analytics/imagery_ask_qwen.py
@@ -15,7 +15,10 @@ ENDPOINT_NAMESPACE = "Kitware"
 ENDPOINT_NAME = "qwen3-5-9b-gguf-ulh"
 MODEL_CARD_URL = "https://huggingface.co/unsloth/Qwen3.5-9B-GGUF"
 SYSTEM_PROMPT = (
-    "You are a geospatial analyst. Answer the following question about the provided image."
+    "You are an urban planning and geospatial analysis expert specializing in "
+    "land use patterns, hydrology, transportation networks, and municipal policy. "
+    "Analyze the provided imagery to answer the user's question. In your answer, "
+    "assume that the user is also a geospatial analyst with the same expertise."
 )
 TOKEN_RANGE = {"min": 1000, "max": 10000, "step": 1000}
 THUMBNAIL_WIDTH = 2000
--- a/uvdat/core/tasks/analytics/imagery_ask_qwen.py
+++ b/uvdat/core/tasks/analytics/imagery_ask_qwen.py
@@ -119,7 +119,7 @@ def imagery_ask_qwen(result_id):
 
     result.write_status("Starting inference endpoint...")
     endpoint = get_inference_endpoint(
-        name=settings.UVDAT_HF_ENDPOINT_NAMES.get("qwen"),
+        name=settings.UVDAT_HF_ENDPOINT_NAMES["qwen"],
         namespace=settings.UVDAT_HF_NAMESPACE,
         token=settings.UVDAT_HF_TOKEN,
     )
--- a/uvdat/core/tasks/analytics/imagery_ask_qwen.py
+++ b/uvdat/core/tasks/analytics/imagery_ask_qwen.py
@@ -1,0 +1,140 @@
+from __future__ import annotations
+
+import base64
+
+from celery import shared_task
+from django.conf import settings
+from django_large_image import utilities
+import large_image
+
+from uvdat.core.models import RasterData, TaskResult
+
+from .analysis_type import AnalysisInputError, AnalysisTask, AnalysisType
+
+ENDPOINT_NAMESPACE = "Kitware"
+ENDPOINT_NAME = "qwen3-5-9b-gguf-ulh"
+MODEL_CARD_URL = "https://huggingface.co/unsloth/Qwen3.5-9B-GGUF"
+PROMPT_PREFIX = (
+    "You are a geospatial analyst. Answer the following question about the provided image."
+)
+TOKEN_RANGE = {"min": 1000, "max": 10000, "step": 1000}
+THUMBNAIL_WIDTH = 2000
+
+
+class ImageryAskQwen(AnalysisType):
+    def __init__(self):
+        super().__init__()
+        self.name = "Imagery: Ask Qwen"
+        self.description = "Select an imagery layer and ask Qwen 3.5 about it."
+        self.details = (
+            "Inferencing with unsloth/Qwen3.5-9B-GGUF provided by a "
+            "Kitware-hosted Huggingface Inference Endpoint. "
+            f"See the model card at {MODEL_CARD_URL}. "
+            f'Prompts will be prefixed with "{PROMPT_PREFIX}".'
+            "Responses may cut off mid-sentence if max_tokens is reached."
+        )
+        self.db_value = "imagery_ask_qwen"
+        self.input_types = {
+            "imagery": "RasterData",
+            "text_prompt": "string",
+            "max_tokens": "number",
+        }
+        self.output_types = {
+            "response": "markdown",
+        }
+        self.attribution = "Unsloth AI, Kitware Inc."
+
+    @classmethod
+    def is_enabled(cls):
+        return settings.UVDAT_ENABLE_IMAGERY_ASK_QWEN and settings.UVDAT_HF_TOKEN
+
+    def get_input_options(self):
+        return {
+            "imagery": RasterData.objects.filter(dataset__category="imagery"),
+            "text_prompt": [],
+            "max_tokens": [TOKEN_RANGE],
+        }
+
+    def validate_inputs(self, inputs):
+        super().validate_inputs(inputs)
+        try:
+            imagery = RasterData.objects.get(id=inputs.get("imagery"))
+        except RasterData.DoesNotExist as e:
+            err_msg = "Imagery raster does not exist."
+            raise AnalysisInputError(err_msg) from e
+        if imagery.dataset.category != "imagery":
+            err_msg = 'Selected raster is not categorized as "imagery".'
+            raise AnalysisInputError(err_msg)
+        max_tokens = int(inputs.get("max_tokens"))
+        if max_tokens < TOKEN_RANGE["min"] or max_tokens > TOKEN_RANGE["max"]:
+            err_msg = f"max_tokens must be between {TOKEN_RANGE['min']} and {TOKEN_RANGE['max']}."
+            raise AnalysisInputError(err_msg)
+
+    def run_task(self, *, project, **inputs):
+        text_prompt = inputs.get("text_prompt")
+        result = TaskResult.objects.create(
+            name=text_prompt[:250],
+            task_type=self.db_value,
+            inputs=inputs,
+            project=project,
+            status="Initializing Task...",
+        )
+        imagery_ask_qwen.delay(result.id)
+        return result
+
+    def finalize(self, result):
+        pass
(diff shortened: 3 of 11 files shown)


uvdat/settings/base.py
```diff
@@ -158,6 +158,7 @@
 }
 
 UVDAT_WEB_URL: str = env.url("DJANGO_UVDAT_WEB_URL").geturl()
+UVDAT_HF_TOKEN: str = env.str("DJANGO_UVDAT_HF_TOKEN", default=None)
```
brianhelba:
```suggestion
UVDAT_HF_TOKEN: str | None = env.str("DJANGO_UVDAT_HF_TOKEN", default=None)
```

uvdat/settings/base.py
```diff
@@ -172,6 +173,7 @@
 UVDAT_ENABLE_UNCERTAINTY_QUANTIFICATION: bool = env.bool(
     "DJANGO_UVDAT_ENABLE_UNCERTAINTY_QUANTIFICATION", default=True
 )
+UVDAT_ENABLE_IMAGERY_ASK_QWEN: bool = env.bool("DJANGO_UVDAT_ENABLE_IMAGERY_ASK_QWEN", default=True)
```
brianhelba, line 179: Do we need this variable at all? Right now, it seems like `settings.UVDAT_HF_TOKEN is None` provides the same signal.

annehaley: I'm just following the same pattern that we have for other tasks; each one has a boolean setting controlled by an environment variable so that the task can be disabled in a given instance. It's true that the HF token also controls this, but there will be other tasks that use the same token soon, and we may want a way to disable one without disabling all HF tasks.

brianhelba: Ah, got it.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,140 @@
...
+            "max_tokens": "number",
+        }
+        self.output_types = {
+            "response": "markdown",
+        }
+        self.attribution = "Unsloth AI, Kitware Inc."
+
+    @classmethod
+    def is_enabled(cls):
+        return settings.UVDAT_ENABLE_IMAGERY_ASK_QWEN and settings.UVDAT_HF_TOKEN
```
brianhelba:
```suggestion
    def is_enabled(cls) -> bool:
        return settings.UVDAT_ENABLE_IMAGERY_ASK_QWEN and settings.UVDAT_HF_TOKEN is not None
```

The return type wasn't always a bool before, as sometimes the value of `settings.UVDAT_HF_TOKEN` could be returned.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,140 @@
...
+    thumbnail_uri = f"data:image/jpeg;base64,{thumbnail_b64}"
+
+    result.write_status("Starting inference endpoint...")
+    endpoint = get_inference_endpoint(
+        name=ENDPOINT_NAME,
+        namespace=ENDPOINT_NAMESPACE,
+        token=settings.UVDAT_HF_TOKEN,
+    )
+    endpoint.resume()
+    endpoint.wait()
```
brianhelba: Without a `timeout`, this could hang indefinitely: https://huggingface.co/docs/huggingface_hub/v1.20.1/en/package_reference/inference_endpoints#huggingface_hub.InferenceEndpoint.wait

We should set some upper bound for waiting, then just accept that it's a failure.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,140 @@
...
+            "role": "user",
+            "content": [
+                {"type": "image_url", "image_url": {"url": thumbnail_uri}},
+                {"type": "text", "text": f"{PROMPT_PREFIX} {text_prompt}"},
+            ],
+        }
+    ]
+
+    result.write_status("Awaiting Qwen's response...")
+    chat = endpoint.client.chat.completions.create(
```
brianhelba:
```suggestion
    chat = endpoint.client.chat_completion(
```

Looks like this is more canonical: https://huggingface.co/docs/huggingface_hub/v1.20.1/en/package_reference/inference_client#huggingface_hub.InferenceClient.chat_completion

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,140 @@
...
+        model="unsloth/Qwen3.5-9B-GGUF",
+        messages=messages,
+        stream=False,
```
brianhelba: This already defaults to `False`: https://huggingface.co/docs/huggingface_hub/v1.20.1/en/package_reference/inference_client#huggingface_hub.InferenceClient.chat_completion.stream

I don't object if you want to express something semantically by leaving it in, but maybe that would warrant a code comment.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,140 @@
...
+
+@shared_task(base=AnalysisTask)
+def imagery_ask_qwen(result_id):
+    # Only available with [tasks] extra
+    from huggingface_hub import get_inference_endpoint  # noqa: PLC0415
+
+    result = TaskResult.objects.get(id=result_id)
+    imagery = RasterData.objects.get(id=result.inputs.get("imagery"))
+    text_prompt = result.inputs.get("text_prompt")
+    max_tokens = int(result.inputs.get("max_tokens"))
```
brianhelba, line 111: Is this determined by the end user? Should we enforce some cap?

annehaley: The accepted range for max tokens is 1k -10k. This is defined at the top of the file and validated in `validate_inputs` (L68-71).

uvdat/core/tasks/analytics/imagery_ask_qwen.py
annehaley: We don't need it to express anything in particular. Removed in [78c87a5](https://github.com/OpenGeoscience/geodatalytics/pull/406/commits/78c87a5be95a402c546e89b2503a032240ff0699).

uvdat/core/tasks/analytics/imagery_ask_qwen.py
annehaley: Changed in [78c87a5](https://github.com/OpenGeoscience/geodatalytics/pull/406/commits/78c87a5be95a402c546e89b2503a032240ff0699).

uvdat/core/tasks/analytics/imagery_ask_qwen.py
annehaley: Added a 5 minute timeout in [78c87a5](https://github.com/OpenGeoscience/geodatalytics/pull/406/commits/78c87a5be95a402c546e89b2503a032240ff0699).

uvdat/core/tasks/analytics/imagery_ask_qwen.py
annehaley: Changed in [78c87a5](https://github.com/OpenGeoscience/geodatalytics/pull/406/commits/78c87a5be95a402c546e89b2503a032240ff0699)

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,147 @@
...
+        result.write_error("Endpoint failed to start in 5 minutes. Try again later.")
+        return
+
+    result.write_status("Sending question to Qwen...")
+    messages = [
+        {
+            "role": "user",
+            "content": [
+                {"type": "image_url", "image_url": {"url": thumbnail_uri}},
+                {"type": "text", "text": f"{PROMPT_PREFIX} {text_prompt}"},
```
manthey: Rather than prefixing the prompt, should we instead use a system message?

If we make this a general endpoint, that is where you would also have more instructions to try to avoid prompt hacking, if we care.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,147 @@
...
+            ],
+        }
+    ]
+
+    result.write_status("Awaiting Qwen's response...")
+    chat = endpoint.client.chat_completion(
+        model="unsloth/Qwen3.5-9B-GGUF",
+        messages=messages,
+        max_tokens=max_tokens,
```
manthey, line 152: Should we be specifying a temperature?

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,147 @@
...
+
+    result = TaskResult.objects.get(id=result_id)
+    imagery = RasterData.objects.get(id=result.inputs.get("imagery"))
+    text_prompt = result.inputs.get("text_prompt")
+    max_tokens = int(result.inputs.get("max_tokens"))
+
+    result.write_status("Encoding imagery...")
+    imagery_path = utilities.field_file_to_local_path(imagery.cloud_optimized_geotiff)
+    src = large_image.open(imagery_path)
+    thumbnail_bytes, _ = src.getThumbnail(THUMBNAIL_WIDTH, encoding="PNG")
```
manthey: We probably want pass the limit to both width and height, or a long narrow image will be too large. And, I think the maximum size image will depend on the model and which projector transformers or ollama uses for it.

pyproject.toml
```diff
@@ -75,6 +75,7 @@ tasks = [
   "pyvips==3.1.1.8.18.1",
   "uvdat-flood-sim[large-image-writer]==1.0.4",
   "xdg-base-dirs==6.0.2",
+  "huggingface-hub>=1.14.0",
```
brianhelba:
```suggestion
  "huggingface-hub==1.14.0",
```

Renovate will try to pin this, so you might as well do it now. I'm not saying this is generally a best practice for Python, but it's how Renovate works currently.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,147 @@
...
+from django.conf import settings
+from django_large_image import utilities
+import large_image
+
+from uvdat.core.models import RasterData, TaskResult
+
+from .analysis_type import AnalysisInputError, AnalysisTask, AnalysisType
+
+ENDPOINT_NAMESPACE = "Kitware"
+ENDPOINT_NAME = "qwen3-5-9b-gguf-ulh"
```
brianhelba: Although I previously suggested that we didn't have to keep this a secret, now that I see it hardcoded here, I'm worried that future open source users or developers will try to use our deployment-specific endpoint (I've seen this happen before).

I think we should make this a deployment-specific env var.

To avoid proliferation of multiple env vars, we could potentially pack this into a single URL.

annehaley: Could you elaborate on what you mean by packing this information into a single URL? My first assumption is that you mean to create a JSON mapping of task names to HF endpoint names and serve it from somewhere else, but I wouldn't know where to put it.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
annehaley: Sure, should we make it a user input or define it as a constant?

uvdat/core/tasks/analytics/imagery_ask_qwen.py
annehaley: Alright, [d70f3ea](https://github.com/OpenGeoscience/geodatalytics/pull/406/commits/d70f3ea249c20033ca16e7350a4e1acf4a7f617a) structures the messages as one system prompt followed by one user prompt. We could make the system prompt more specific if you have any ideas on what to include.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
manthey: I don't know.  Locally, I ran the same walkable question a few times and once qwen decided we were looking at Hartford rather than Boston.  Let me see empirically how much it affects the model we've picked.

annehaley: We discussed this in our internal meeting and decided that the model's default temperature is usually most appropriate. In the case of this model, @manthey found that different temperature values do not yield noticeably different results, so we'll keep the default temperature for now. As we add more models for other tasks, we'll evaluate the temperature variance on those accordingly.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
annehaley: Alright, [8a4ab65](https://github.com/OpenGeoscience/geodatalytics/pull/406/commits/8a4ab65001b2d9e1c29f0175fa899408951fc34e) passes the same max size constant for both width and height. I increased the size to 4000 because that worked fine when I tried it. But with a size of 8000, I got `Client error '413 Payload Too Large'`.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
annehaley: [3511794](https://github.com/OpenGeoscience/geodatalytics/pull/406/commits/3511794e9118130260b01df399cfca8a0294914f) updates the system prompt to "You are an urban planning and geospatial analysis expert specializing in land use patterns, hydrology, transportation networks, and municipal policy. Analyze the provided imagery to answer the user's question. In your answer, assume that the user is also a geospatial analyst with the same expertise."

uvdat/core/tasks/analytics/imagery_ask_qwen.py
annehaley: @brianhelba As per our discussion over Meet, [1ef13dc](https://github.com/OpenGeoscience/geodatalytics/pull/406/commits/1ef13dc313c881414d47115724c23a8f00329157) defines these constants as environment variables instead. The namespace will not change as we add more inference endpoints, but the endpoint names will be different. Endpoint names will be specified as a dictionary; right now that dict just contains the qwen endpoint name, but we will add more with more tasks.

uvdat/core/tasks/analytics/imagery_ask_qwen.py
```diff
@@ -0,0 +1,161 @@
...
+    result.write_status("Encoding imagery...")
+    imagery_path = utilities.field_file_to_local_path(imagery.cloud_optimized_geotiff)
+    src = large_image.open(imagery_path)
+    thumbnail_bytes, _ = src.getThumbnail(THUMBNAIL_SIZE, THUMBNAIL_SIZE, encoding="PNG")
+    thumbnail_b64 = base64.b64encode(thumbnail_bytes).decode("utf-8")
+    thumbnail_uri = f"data:image/jpeg;base64,{thumbnail_b64}"
+
+    result.write_status("Starting inference endpoint...")
+    endpoint = get_inference_endpoint(
+        name=settings.UVDAT_HF_ENDPOINT_NAMES.get("qwen"),
```
brianhelba:
```suggestion
        name=settings.UVDAT_HF_ENDPOINT_NAMES["qwen"],
```
At this point, a missing value should actually trigger an error.

terraform/django.tf
```diff
@@ -20,9 +20,14 @@ module "django" {
...
 
   additional_django_vars = {
-    DJANGO_UVDAT_WEB_URL          = "https://www.geodatalytics.kitware.com/"
-    DJANGO_DATABASE_POOL_MAX_SIZE = "12"
-    DJANGO_SENTRY_DSN             = "https://5302701c88f1fa6ec056e0c269071191@o267860.ingest.us.sentry.io/4510620385804288"
+    DJANGO_UVDAT_WEB_URL           = "https://www.geodatalytics.kitware.com/"
+    DJANGO_DATABASE_POOL_MAX_SIZE  = "12"
+    DJANGO_SENTRY_DSN              = "https://5302701c88f1fa6ec056e0c269071191@o267860.ingest.us.sentry.io/4510620385804288"
+    DJANGO_UVDAT_HF_NAMESPACE      = "Kitware"
+    DJANGO_UVDAT_HF_ENDPOINT_NAMES = "qwen=qwen3-5-9b-gguf-ulh,"
```
brianhelba, line 27: If you end up with any more of these, you can use some Terraform syntax like:
```hcl
locals {
  huggingface_endpoint_names = {
    qwen = "qwen3-5-9b-gguf-ulh"
  }
}

module "django" {
  ...
  additional_django_vars = {
    ...
    DJANGO_UVDAT_HF_ENDPOINT_NAMES = join(",", [for k, v in local.huggingface_endpoint_names : "${k}=${v}"])
  }
}
```

Maybe it's overkill for now, but worth remembering in the future.

brianhelba (commented):
2 minor suggestions inline, otherwise looks good.

terraform/django.tf
annehaley: Ok, thanks. I agree that we can use this in the future when we add more endpoints to the mapping.

brianhelba (approved):

manthey (approved):

PR #407, opened 2026-06-25
Redis "volatile-lru" cache

### LINKED ISSUE #398: Set Redis Max Memory Policy

annehaley:
This PR sets the large image tile cache to use redis with the "volatile-lru" max memory policy. Cached tiles have a TTL of 4 weeks.

See [x](https://github.com/OpenGeoscience/geodatalytics/pull/400) and [x](https://github.com/girder/large_image/pull/2092). Resolves #398.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,6 +48,7 @@ services:
 
   redis:
     image: redis:alpine
+    command: ["redis-server", "--maxmemory", "4096MB", "--maxmemory-policy", "volatile-lru"]
     healthcheck:
       test: ["CMD", "redis-cli", "ping"]
       start_period: 30s
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,7 +48,7 @@ services:
 
   redis:
     image: redis:alpine
-    command: ["redis-server", "--maxmemory", "4096MB", "--maxmemory-policy", "volatile-lru"]
+    command: ["redis-server", "--maxmemory", "4GB", "--maxmemory-policy", "volatile-lru"]
     healthcheck:
       test: ["CMD", "redis-cli", "ping"]
       start_period: 30s
--- a/uvdat/settings/base.py
+++ b/uvdat/settings/base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 import logging
 from pathlib import Path
 from typing import Any
@@ -155,6 +156,23 @@ CHANNEL_LAYERS: dict[str, dict[str, Any]] = {
             ]
         },
     }
+}
+
+# Large image cache with Redis
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+    },
+    "tiles": {
+        "BACKEND": "django.core.cache.backends.redis.RedisCache",
+        "LOCATION": env.url("DJANGO_REDIS_URL").geturl(),
+        "OPTIONS": {
+            # Use database /2 for the tile cache,
+            # in case other services use /0 in the future
+            "db": "2",
+        },
+        "TIMEOUT": int(datetime.timedelta(weeks=4).total_seconds()),
+    },
 }
 
 UVDAT_WEB_URL: str = env.url("DJANGO_UVDAT_WEB_URL").geturl()
--- a/uvdat/settings/base.py
+++ b/uvdat/settings/base.py
@@ -142,13 +142,14 @@ CORS_ALLOWED_ORIGIN_REGEXES: list[str] = env.list(
 )
 
 # django-channels with Redis
+_redis_url = env.url("DJANGO_REDIS_URL").geturl()
 CHANNEL_LAYERS: dict[str, dict[str, Any]] = {
     "default": {
         "BACKEND": "channels_redis.core.RedisChannelLayer",
         "CONFIG": {
             "hosts": [
                 {
-                    "address": env.url("DJANGO_REDIS_URL").geturl(),
+                    "address": _redis_url,
                     # Use database /1 for Channels backend,
                     # in case other services use /0 in the future
                     "db": 1,
@@ -165,12 +166,13 @@ CACHES = {
     },
     "tiles": {
         "BACKEND": "django.core.cache.backends.redis.RedisCache",
-        "LOCATION": env.url("DJANGO_REDIS_URL").geturl(),
+        "LOCATION": _redis_url,
         "OPTIONS": {
             # Use database /2 for the tile cache,
             # in case other services use /0 in the future
             "db": "2",
         },
+        # TTL of 4 weeks (an arbitrarily long time)
         "TIMEOUT": int(datetime.timedelta(weeks=4).total_seconds()),
     },
 }
--- a/terraform/django.tf
+++ b/terraform/django.tf
@@ -45,6 +45,10 @@ module "django" {
 resource "heroku_addon" "redis" {
   app_id = module.django.heroku_app_id
   plan   = "heroku-redis:mini"
+
+  config = {
+    maxmemory_policy = "volatile-lru"
+  }
 }
 
 output "dns_nameservers" {


manthey (approved):

brianhelba:
Why not just do this through the Django cache framework, which is the default for `django-large-image`?

annehaley:
> Why not just do this through the Django cache framework, which is the default for `django-large-image`?

We're currently using the Django cache, but in production when a user tries to visualize the boston orthoimagery layer for our example use cases, the web service with Django runs out of memory: `Process running mem=895M(122.6%)  Error R14 (Memory quota exceeded)`.

brianhelba:
I see that after #404, the Django cache is being used, but it was never actually configured properly.

Per [my suggestion](https://github.com/OpenGeoscience/geodatalytics/pull/400#issuecomment-4635052276), #404 needed to add something like:
```python
LARGE_IMAGE_CACHE_NAME = "tiles"

CACHES = {
    "tiles": {
        "BACKEND": "django.core.cache.backends.redis.RedisCache",
        "LOCATION": env.url("DJANGO_REDIS_URL").geturl(),
        "OPTIONS": {
            # Use database /2 for the tile cache,
            # in case other services use /0 in the future
            "db": "2",
        },
        # An arbitrarily long time
        "TIMEOUT": int(datetime.timedelta(weeks=4).total_seconds()),
    }
}
```

If we need to balance other cache considerations in the future, the configuration [is well-documented by Django](https://docs.djangoproject.com/en/6.0/topics/cache/#redis) and we have all the local knobs (including the actual `TIMEOUT`) to adjust things.

uvdat/settings/base.py
```diff
@@ -157,6 +157,10 @@
     }
 }
 
+# Large image cache with Redis
+LARGE_IMAGE_CACHE_BACKEND = "redis"
+LARGE_IMAGE_CACHE_REDIS_URL = env.url("DJANGO_REDIS_URL").geturl()
```
brianhelba: Where do we control the Redis database number here?

We're using Redis for other things, so I think we should have local control of this.

annehaley:
@brianhelba I tried the above suggestion locally and ran into some problems. First, if I set `CACHES` directly like that, I get an error about how Django needs a "default" cache. To resolve that, I added the "default" entry like this:
```
CACHES = {
    "default": {
        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
    },
    "tiles": {
        "BACKEND": "django.core.cache.backends.redis.RedisCache",
        "LOCATION": env.url("DJANGO_REDIS_URL").geturl(),
        "OPTIONS": {
            # Use database /2 for the tile cache,
            # in case other services use /0 in the future
            "db": "2",
        },
        "TIMEOUT": int(datetime.timedelta(weeks=4).total_seconds()),
    }
}
```

But the next problem is more prohibitive. After that change, it appeared to work properly, but when I performed minimal load testing, I got consistent crashes. I was watching the logs from the redis container as I loaded our aerial imagery layer on the map. It appeared to be working at first; tile fetching after the first load was snappy and the redis logs reflected the cache hits. But after a few refreshes loading the same tiles on the map, I noticed the following behavior:

1. The django logs show an error about how the web request process "took too long to shut down and was killed".
2. In the web client, tiles are not shown on the map and a generic error dialog is shown (expected behavior for a failed request).
3. The redis logs stop printing output.
4. The performance of my whole machine slows significantly, to the point of my mouse cursor lagging and my Chrome crashing.
5. The redis container shuts down and needs to be restarted manually. The redis logs do not record any errors, so this leads me to believe that it was OOM killed.

brianhelba:
Just to ensure the problem is truly with the `"tiles"` cache and Redis, I'd set the `"default"` cache to `django.core.cache.backends.dummy.DummyCache`: https://docs.djangoproject.com/en/6.0/topics/cache/#dummy-caching-for-development

annehaley:
When I run geodatalytics not in a devcontainer (classic `docker compose up` style), I can get to an error message (instead of just a logless crash): `redis.exceptions.OutOfMemoryError: command not allowed when used memory > 'maxmemory'`

Changing the "default" cache to a `DummyCache` doesn't change anything. I get the same behavior for both.

If I increase the `maxmemory` value in `docker-compose.yml` (L51) to something large like "40GB", the tiles load successfully and no error is raised. But even "30GB" is still too small and will raise errors. 

My expectation was that redis would handle reaching its memory limit gracefully by evicting old keys for tiles (for which we've set a default TTL). I don't know why it would be raising an error like this when the `maxmemory-policy` is set to `volatile-lru`.

One additional thing to note is that when the redis container first starts, it takes about 30-60 seconds printing out a bunch of logs that all say `* <search> Draining workers thread pool with threshold 100`.  If any tile requests are made during this time, I get this error: `redis.exceptions.BusyLoadingError: Redis is loading the dataset in memory`.

manthey:
I tested `volatile-lru` behavior with Histomics (not Django) and it properly evicted tiles as expected with `voltaile-lru`, even for small sizes like 10 MB.

brianhelba:
@annehaley I think your issue may be because you have tile data in Redis from before the addition of the TTL, which will never be evicted, even with the new policy.

Try fully resetting Redis by dropping the volume:
```
docker compose stop
docker compose down -v redis
```

brianhelba (changes requested):
To sum up my feedback, I think you should use Django's cache framework, [per my comment here](https://github.com/OpenGeoscience/geodatalytics/pull/407#issuecomment-4810995368) (adding the `default` entry too).

If you really want to go with this approach instead, I'd suggest trying to deal with the database number issue [in this comment](https://github.com/OpenGeoscience/geodatalytics/pull/407/changes#r3482476882).

annehaley:
@brianhelba You were right, the OOM kills were due to the fact that my redis volume had previous cache entries stored without a TTL. After clearing that volume I was able to use the redis cache with your suggested changes. I've rebased this PR with that approach.

uvdat/settings/base.py
```diff
@@ -157,6 +158,23 @@
...
 }
 
+# Large image cache with Redis
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+    },
+    "tiles": {
+        "BACKEND": "django.core.cache.backends.redis.RedisCache",
+        "LOCATION": env.url("DJANGO_REDIS_URL").geturl(),
```
brianhelba: Not a big deal, but maybe pull this out to a local and private variable (like `_redis_url`), since it's used both here and in `CHANNEL_LAYERS`.

uvdat/settings/base.py
```diff
@@ -157,6 +158,23 @@
     }
 }
 
+# Large image cache with Redis
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
```
brianhelba, line 165: If this is ever actually used, it should be changed to a more robust alternative. In production, this will result in isolated caches for each worker, which won't allow sharing of cached data and will increase the memory usage of each worker.

Right now, the only thing that I believe may use this is Django-Allauth for a few rate-limiting / OAuth2 flow things.

uvdat/settings/base.py
```diff
@@ -157,6 +158,23 @@
...
+    },
+    "tiles": {
+        "BACKEND": "django.core.cache.backends.redis.RedisCache",
+        "LOCATION": env.url("DJANGO_REDIS_URL").geturl(),
+        "OPTIONS": {
+            # Use database /2 for the tile cache,
+            # in case other services use /0 in the future
+            "db": "2",
+        },
+        "TIMEOUT": int(datetime.timedelta(weeks=4).total_seconds()),
```
brianhelba, line 176: Maybe it's obvious, but you could add a comment that our 4 weeks wasn't precisely chosen, but is just an "arbitrarily long time". I don't feel strongly.

docker-compose.yml
```diff
@@ -48,6 +48,7 @@ services:
 
   redis:
     image: redis:alpine
+    command: ["redis-server", "--maxmemory", "4096MB", "--maxmemory-policy", "volatile-lru"]
```
brianhelba:
```suggestion
    command: ["redis-server", "--maxmemory", "4GB", "--maxmemory-policy", "volatile-lru"]
```
Any unit is supported: https://raw.githubusercontent.com/redis/redis/8.8/redis.conf

brianhelba (approved):
Some minor code style issues that you might want to clean up, but LGTM structurally.

uvdat/settings/base.py
annehaley: The `LocMemCache` is already used for the "default" cache when we do not specify the `CACHES` config (https://docs.djangoproject.com/en/6.0/topics/cache/#local-memory-caching). We can address this separately if it ends up being a problem.

docker-compose.yml
annehaley: Changed in [098a52c](https://github.com/OpenGeoscience/geodatalytics/pull/407/commits/098a52c5eedbb38d6cf90faa17f0b0e2e951c1f8)

uvdat/settings/base.py
annehaley: Added comment in [098a52c](https://github.com/OpenGeoscience/geodatalytics/pull/407/commits/098a52c5eedbb38d6cf90faa17f0b0e2e951c1f8)

uvdat/settings/base.py
annehaley: Changed in [098a52c](https://github.com/OpenGeoscience/geodatalytics/pull/407/commits/098a52c5eedbb38d6cf90faa17f0b0e2e951c1f8)

uvdat/settings/base.py
brianhelba: Understood and agreed. Just a warning.

brianhelba (approved):

brianhelba:
I missed the fact that this isn't setting `LARGE_IMAGE_CACHE_NAME`. 😓

annehaley:
Good catch @brianhelba. I made https://github.com/OpenGeoscience/geodatalytics/pull/425 as a follow-up.

PR #414, opened 2026-06-30
Redirect to www

annehaley:
This PR adds Terraform configuration to redirect requests from geodatalytics.kitware.com to www.geodatalytics.kitware.com.
Copied from [this file from Mixtape](https://github.com/Kitware/mixtape/blob/master/terraform/apex_redirect.tf).


brianhelba (approved):
:shipit:

Be sure to approve the apply operation in TF Cloud after merging.

PR #415, opened 2026-06-30
Decrease thumbnail size for qwen imagery task

annehaley:
Now that #406 has been merged, I was able to test the behavior in production. My first attempt failed with an OOM kill during the step where we encode the imagery thumbnail (killed before it ever reached out to the inference endpoint). I made a temporary local change to decrease the thumbnail size constant from 4000 to 2000 and tried again, and it worked. 

I had increased the thumbnail size to 4000 when we found that the inference endpoint could handle it, but now I find that the limiting factor is the memory of the celery worker, not the memory of the inference endpoint.


manthey:
The model we picked will handle image up to 4096^2 pixels (it doesn't have to be square); anything larger gets scaled down by a pre-processing step of some sort, so probably the endpoint would allow even bigger images through.  (This does nothing to help the worker's memory use)

Issue #410, opened 2026-06-26
Write user documentation


annehaley:
We should write and deploy user documentation, including screenshots and explanations of each part of the user interface, as well as details for advanced use cases. The documentation should include the following sections:

- [x] Introduction
- [x] Project Management
    - [x] Dataset Upload
- [x] Map controls
    - [x] Controls bar
- [x] How to use Panels
    - [x] Datasets Panel
    - [x] Selected Layers Panel
    - [x] Legend Panel
    - [x] Charts Panel
    - [x] Networks Panel
    - [x] AI & Analytics Panel
- [x] Advanced Use Cases
    - [x] API usage
    - [x] Custom Analytics Tasks

Optionally, it would be cool if we could add a CI task to automatically recapture screenshots for this documentation so we can avoid letting it get out of date due to UI changes.

annehaley:
Drafts are written [here](https://docs.google.com/document/d/1RPNoYeE7pA8wS81tiJY0kg8ptfsfuToQO_J5907KGEY/edit?usp=sharing)

PR #418, opened 2026-07-02
UI Adjustments 1

### LINKED ISSUE #378: Drag divider between 'Selected Layer' and 'datasets' accordions disappear.
When the 'selected layers' accordion is collapsed, the drag divider above it disappears and doesn't show even when the accordion is expanded again. It appears again when the dataset accordion is collapsed.
Expected behavior: the divider doesn't disappear unless there is only one accordion in the left panel.

Image

annehaley:
In the process of writing user documentation for #410, I noticed several places where the UI could use some small fixes. This PR is the first batch; I assume there will be more as I continue to write the docs.

Resolves #378.


web/src/components/sidebars/LayersPanel.vue
```diff
@@ -112,15 +112,13 @@ function setLayerActive(layer: Layer, active: boolean) {
...
         />
         <v-checkbox-btn
           v-if="!isComparing"
-          :model-value="
-            layerStore.selectedLayers.every((l: Layer) => l.visible)
-          "
+          :model-value="filteredLayers.every((l: Layer) => l.visible)"
           style="display: inline"
-          @click="setVisibility(layerStore.selectedLayers, !allLayersVisible)"
+          @click="setVisibility(filteredLayers, !allLayersVisible)"
```
BryonLewis: This is correct in that it should use the search filter.  The question is if the toggling of the layers should be be based on `allLayersVisible` or should it be based on `allFilteredLayersVisible` - a new computed property for only the filtered set of data?

web/src/components/sidebars/FloatingPanel.vue
```diff
@@ -211,7 +211,7 @@ function panelUpdated() {
       </v-card>
     </div>
     <div
-      v-if="!panel.position && !props.bottom"
+      v-if="!panel.position && !panel.collapsed"
```
BryonLewis, line 213: the bottom prop is still passed in inside of calls to `<floatingPanel>` and is defined in the props for this component but unsued now?

web/src/App.vue
```diff
@@ -118,4 +118,37 @@ scrollbar changes disables default styling and rounding so corners appear square
...
+  color: rgb(var(--v-theme-success));
+}
+
+.v-number-input .v-field__field {
+  flex-direction: column;
+}
+
+.v-input {
+  --v-input-control-height: 32px;
+}
```
BryonLewis, line 132: Could some of this be done with Vuetify settings like global defaults instead of utilizing a higher level css override? (specifically for the v-icons).
https://vuetifyjs.com/en/features/global-configuration/#contextual-defaults

Something like:
```
 defaults: {
    VIcon: {
      color: "secondary-text",
    },
  },
```
then removing these additional CSS fields (maybe keeping the opacity) should hopefully make the icons respect the color="success" and other fields.  Be sure to check the previous icons, I'm not a hundred percent sure that this will work.

BryonLewis (changes requested):
Just some minor questions on some changes.

web/src/components/sidebars/FloatingPanel.vue
annehaley: Good point. I removed that unused prop in [db0ecb3](https://github.com/OpenGeoscience/geodatalytics/pull/418/commits/db0ecb31f350236953e44eb79a35f86dfd16676f)

web/src/components/sidebars/LayersPanel.vue
annehaley: Changed in [91358e0](https://github.com/OpenGeoscience/geodatalytics/pull/418/commits/91358e034a42570665065b6f4c51bdc7276af452)

web/src/App.vue
annehaley: I did my best to accomplish this with [41d53a2](https://github.com/OpenGeoscience/geodatalytics/pull/418/commits/41d53a2b13cd6a9f68c3f2faf95d6f53d98a9c7c). My goal was to set the icon color to `secondary-text` unless the icon is inside a button with a color applied. The first default matches your suggestion; that takes care of icons not inside buttons. Another default overrides that rule for icons inside buttons, which should inherit the color from the button. Then I added another default such that buttons have a color of  "transparent" and a css rule that declares that icons inside transparent buttons should be `secondary-text` colored. So buttons with a special color applied like `primary` or `success` will have icons colored to match the button text, but all other icons will default to `secondary-text`.

One additional change was necessary for the compare layers icon, which has the color set programatically. This is still a complex solution, but it involves fewer top-level css overrides. Lmk your thoughts.

BryonLewis (approved):
Pulled and did a quick test, everything I saw looked good to me.

PR #421, opened 2026-07-20
Store creator for new TaskResults

### LINKED ISSUE #413: Add `creator` field to `TaskResult`
We will need to start tracking which user created a given `TaskResult`, both for analytics tasks and for dataset conversion tasks immediately following dataset uploads. This will be helpful to better understand usage patterns. This is also a prerequisite for implementing a paid tier for either feature.

annehaley:
Resolves #413


uvdat/core/models/task_result.py
```diff
@@ -29,6 +29,9 @@ class TaskResult(models.Model):
     error = models.TextField(blank=True, default="")
     created = models.DateTimeField(auto_now_add=True, editable=False)
     completed = models.DateTimeField(null=True)
+    creator = models.ForeignKey(
+        User, related_name="task_results", null=True, on_delete=models.CASCADE
```
BryonLewis: are you sure want this to cascade on a user being deleted?
It's valid but I just want to make sure that you're okay with TaskResults being cleared if a user is deleted even though the project still exists?

uvdat/core/models/dataset.py
```diff
@@ -107,6 +107,7 @@ def spawn_conversion_task(
                     "region_options": region_options,
                 },
                 status="Initializing task...",
+                creator=self.owner(),
```
BryonLewis, line 111: This is a minor nit and maybe won't matter.  This is setting the self.owner() to be the creator.  The `spawn_conversion_task` can be called by the `POST /dataset/{id}/convert` which can be called by other user's not necessariyl the owner if I'm correct.  Should  the spawn_conversion_task pass the owner in for this, so it could use the request user when in the POST endpoint but in other ingest commands it would use the dataset owner?

It's minor because I know in all practical uses it's the owner that would be spawing the conversion task, but the API endpoint is there for other users to do it.

BryonLewis (approved):
Some minor comments/questions.  I'll let you determine if they need to be addressed or not.  I'm fine not addressing them but just wanted to bring them up if they weren't considered.

uvdat/core/models/task_result.py
annehaley: Good point, `SET_NULL` is probably more appropriate. Changed in [dff1948](https://github.com/OpenGeoscience/geodatalytics/pull/421/commits/dff19486338c0ace026c7de81b13cf1f7f6fe6e1)

uvdat/core/models/dataset.py
annehaley: Ok, I added a few lines to overwrite the result's creator in that conversion endpoint ([bbabfea](https://github.com/OpenGeoscience/geodatalytics/pull/421/commits/bbabfeac39ccdee145677f4fb4a4f4054853ccd8)). The default will still be the dataset owner if the conversion task is spawned from anywhere else.

BryonLewis (approved):
👍

PR #422, opened 2026-07-20
Fix validation logic for dataset upload form

annehaley:
While writing the documentation for dataset upload, I found a bug in the validation logic. The max upload size validation rule never returns `undefined`; it returns a string error message or `true`. In the definition of the computed `valid` variable, we had checked for `undefined` values. This PR changes that logic to check whether an error message string was returned.

### DIFF
--- a/web/src/components/projects/DatasetUpload.vue
+++ b/web/src/components/projects/DatasetUpload.vue
@@ -75,7 +75,7 @@ const valid = computed(() => {
       return (
         l.name &&
         l.files.length &&
-        fileUploadRules.every((rule) => rule(l.files) === undefined) &&
+        fileUploadRules.every((rule) => rule(l.files) === true) &&
         (l.frame_method === "single" || l.frame_property)
       );
     })


web/src/components/projects/DatasetUpload.vue
```diff
@@ -75,7 +75,7 @@ const valid = computed(() => {
       return (
         l.name &&
         l.files.length &&
-        fileUploadRules.every((rule) => rule(l.files) === undefined) &&
+        fileUploadRules.every((rule) => typeof rule(l.files) !== "string") &&
```
BryonLewis: Sorry again this is minor. I know that your current fileUploadRules is a single rule that returns and error string or a true value.  

Should the check be more inline with vuetify to check for explict true (`=== true`) instead of if the `typeof rule(l.files) !== "string"`?

https://vuetifyjs.com/en/api/v-validation/#props-rules - the documentation supports `true`, `false`, `string`.  So in the future if you have more rules and they return `'false` (meaning not valid) they could pass this check. 

I do understand that this is inside of a computed function that will ultimately evaluate to either `true`/`false` and that it isn't attached to vuetify's `:rule` directive but wondering if the change would help for consistency.

BryonLewis (changes requested):
relatively minor question about using explict `true` vs not a string check on rule validation.

This kind of brings up something else, would this be cleaner to utilize a `v-form` with a `v-model="isFormValid"` and use that variable to disable the submit button and || it with submitting.  Then it checks all of the rules in the v-form automatically without needing the computed function combining previous rules?

https://vuetifyjs.com/en/components/forms/#usage

https://vuetifyjs.com/en/components/forms/#validation-state


I would understand not wanting to make a large a change to to this component.

web/src/components/projects/DatasetUpload.vue
annehaley: Yeah I agree that `=== true` is more consistent and readable. Changed in [9542e69](https://github.com/OpenGeoscience/geodatalytics/pull/422/commits/9542e69fb1e028fd08141cad485e0bca395ddd14)

annehaley:
It's true that a `v-form` can help to consolidate some of the validation logic, but when I tried this out, I couldn't find a clean way to implement the last part of the current `valid` computed property. We need to ensure that for every layer configuration, `(l.frame_method === "single" || l.frame_property)`. 

The frame method and frame property are controlled with `v-btn-toggle` controls, which do not accept a `rules` prop. We could create a hidden input control here, but I think that would increase complexity beyond what it's worth.

BryonLewis (approved):
Thanks for the explanation about the vbtn-toggle and the interaction with the  frame stuff.  I didn't think of that when I quickly looked at the component.  Everything looks good.

PR #423, opened 2026-07-20
Fix screenshot map-only behavior

annehaley:
As I was writing the documentation for the controls bar, I found this bug. The `takeScreenshot` function is used for the screenshot menu AND for creating the thumbnail for a new saved view. That function used `mapOnly` to determine whether to include sidebar elements in the screenshot, but `mapOnly` should only apply to the screenshots menu. If `mapOnly` was checked in the screenshot menu, the thumbnail for a new saved view would not include the sidebars. This PR adjusts that behavior such that sidebars are always included in new view thumbnails.

### DIFF
--- a/web/src/components/ControlsBar.vue
+++ b/web/src/components/ControlsBar.vue
@@ -198,7 +198,7 @@ async function fitMap() {
   loadingBounds.value = false;
 }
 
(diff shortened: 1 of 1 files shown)


BryonLewis (approved):
LGTM

Issue #151, opened 2025-06-10
Create more concrete distinction between duplicate layers in client


jjnesbitt:
Since we have the ability to add a layer twice on the client, we should rigorously define what each of those added layers is, and distinguish them from backend layers (for which there is only one instance). I was thinking of naming each layer that is added to the map something like `LayerInstance`. These could have a unique identifier that is essentially just the layerId, combined with it's copy number (`<layerId>.<copyId>`).

Currently, the distinguishing of different layer copies is done sort of ad hoc everywhere it's needed. This change would standardize that process.

PR #424, opened 2026-07-21
Fix Legend Panel Bug & Style

annehaley:
As I was writing the documentation for the legend panel, I found this bug. If two copies of the same layer are added to the map, the eye icon for one copy will toggle the visibility of both copies. We had two slightly different `setVisibility` functions, one for the selected layers panel and one for the legend panel. To resolve the bug I consolidated the logic in a new `layerStore` function, `setLayerVisibility`. Both panels now use the same function to toggle layer visibility.

Edit: I also found a few more things to fix:
1. Use bold text to emphasize substyle names
2. Apply a maximum precision to colormap range labels

### DIFF
--- a/web/src/components/sidebars/ColormapPreview.vue
+++ b/web/src/components/sidebars/ColormapPreview.vue
@@ -67,9 +67,9 @@ watch(
 
 <template>
   <div class="d-flex" style="column-gap: 5px; width: 100%">
-    <span v-if="props.range">{{ props.range[0] }}</span>
+    <span v-if="props.range">{{ props.range[0]?.toPrecision(3) }}</span>
     <canvas ref="canvas" class="colormap-canvas"></canvas>
-    <span v-if="props.range">{{ props.range[1] }}</span>
+    <span v-if="props.range">{{ props.range[1]?.toPrecision(3) }}</span>
   </div>
 </template>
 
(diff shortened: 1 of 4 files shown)


BryonLewis (approved):
This PR is great for fixing that issue.  Have you completed an audit of `layer.id` access elsewhere in the application to see where it could cause conflicts if there are two layers with the same id?

I think in LegenPanel.vue itself there is a v-for key that is using layer.id which would cause reactivity/updating issues with copies.  There may be places elwhere in the code similar or that utilize the concept of layer.id being a unique identifier within the client while that may not be true.  I would check where layers are stored for settings, or compared for visiblity/selection purposes.

The above could be a separate PR to not mix in more changes with the direct Legend and styling changes.

annehaley:
> Have you completed an audit of `layer.id` access elsewhere in the application to see where it could cause conflicts if there are two layers with the same id?

#151 was created in regards to this concern, but it's been sitting unaddressed for a while now. It could involve a large client refactor, so we've been unintentionally avoiding it, instead opting for small bug fixes like this one. Hopefully we can address it directly soon.

BryonLewis:
> #151 was created in regards to this concern, but it's been sitting unaddressed for a while now. It could involve a large client refactor, so we've been unintentionally avoiding it, instead opting for small bug fixes like this one. Hopefully we can address it directly soon.

Thanks, I should have though to do a search of the issues.

PR #426, opened 2026-07-26
Sentinel download and rgb fixes

BryonLewis:
extracted out of #397 this PR's to make it a bit easier for Review of Frame Preview PR's

### Sentinel downloader (`scripts/sentinelDownload/`)
- Supports creating multiframe single-image Sentinel raster datasets via `--single-file` (uses `gdal_translate` to append frames as subdatasets)
- Additional CLI arguments for date range, cloud cover, output naming, and clip size
- Generates an ingest JSON alongside clipped GeoTIFFs for direct import into Geoodatalytics

### Dataset / RGB fixes
- **Default `band:1` injection removed** when no `source_filter` is supplied — this was preventing Boston ortho imagery from being recognized as RGB
- **`band:1` is still applied when explicitly defined** in the style spec
- **Multiframe rasters no longer render as grayscale** — `frame` and `band` are extracted from the style JSON and passed as separate query parameters (outside `style`) so django-large-image can auto-detect RGB imagery
- Stop embedding frame and band inside the large-image style JSON. Those keys select which raster slice to read, so they are now sent as top-level tile query params.
- Treat legacy ingest { "band": 1 } as “no band filter,” so RGB imagery is not forced into single-band grayscale. Explicit non-default band / frame values from source_filters still apply.
- Centralize tile URL building in buildRasterTileQueryParams and use it from both the map and style stores so source creation and style updates stay consistent.

### DIFF
--- a/web/src/store/style.ts
+++ b/web/src/store/style.ts
@@ -89,6 +89,51 @@ export function colormapMarkersSubsample(
   return markers;
 }
 
+// frame/band select which slice to read; they must be flat query params, not in style JSON
+const RASTER_SOURCE_FILTER_KEYS = new Set(["frame", "band"]);
+
+// Ingest defaults missing source_filters to { band: 1 }; that is not a real band selection.
+function isDefaultBandSourceFilter(
+  sourceFilters: Record<string, unknown>,
+): boolean {
+  const keys = Object.keys(sourceFilters);
+  return (
+    keys.length === 1 &&
+    keys[0] === "band" &&
+    (sourceFilters.band === 1 || sourceFilters.band === "1")
+  );
+}
+
+function sourceFiltersToStyleFilters(
+  sourceFilters: Record<string, unknown> | undefined,
+): StyleFilter[] {
+  if (!sourceFilters || !Object.keys(sourceFilters).length) return [];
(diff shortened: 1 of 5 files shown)


web/src/store/style.ts
```diff
@@ -89,6 +89,51 @@ export function colormapMarkersSubsample(
   return markers;
 }
 
+// frame/band select which slice to read; they must be flat query params, not in style JSON
+const RASTER_SOURCE_FILTER_KEYS = new Set(["frame", "band"]);
+
+// Ingest defaults missing source_filters to { band: 1 }; that is not a real band selection.
```
annehaley: I think we don't need to support previously ingested frames with this filter value; we could just manually modify existing frames to remove that wrong default filter. That filter never should have been applied, so I think it's cleaner to fix existing objects than to work around the presence of that wrong value.

BryonLewis: Yeah that was a bit premature to handle.  I didn't know if I should create a migration because I can't really know the intent behind band: 1 and if it was a default versus and intended filter for only the first band.

I've removed the function and the checking in: https://github.com/OpenGeoscience/geodatalytics/pull/426/commits/d8e53e7cada3f36551dc9a4ab714a9941c2a8915

annehaley (approved):

PR #427, opened 2026-07-26
Fingerprint frame layer previews

BryonLewis:
Modifies #397  to not have a LayerStyle connection and rely only the fingerprint and layer foreign key for indentifying a preview.

## New Models
- **`RasterFramePreview`**: caches one rendered PNG per `(layer_frame, style_fingerprint)`. Fields include `style_fingerprint` (sha256 of `raster_style_params`), `raster_style_params` (render snapshot), `status` (`creating` / `regenerating` / `complete` / `failed`), S3 `image`, plus `width` / `height` / `bounds`. Unique constraint on `(layer_frame, style_fingerprint)`. Deleting a row also deletes its S3 image.
- **`LayerStyle.raster_style_params`**: JSONField holding the client-computed django-large-image style query used for tiles and previews.
- **`Layer` helpers**: `raster_frames()` and `is_multiframe_raster()` (previews only apply when a layer has more than one raster frame).
- Migration: `0025_raster_frame_preview.py`.

## Frame Preview Workflow
Previews are content-addressed by a fingerprint of `raster_style_params` (`params_fingerprint` / `style_fingerprint`).
1. **Enqueue** via `invalidate_and_enqueue_layer_previews`:
   - Skip non-multiframe layers.
   - Skip if every frame already has a complete preview for that fingerprint.
   - Skip if a `frame_preview` `TaskResult` is already in flight for `(layer_id, fingerprint)`.
   - Otherwise upsert preview rows (`creating` / `regenerating`), clear old images, create a `TaskResult` (`task_type="frame_preview"`), and run `generate_frame_previews`.
2. **Generation** (`generate_frame_previews`): render a styled PNG thumbnail per frame with large-image, store on the preview row, mark `complete` (or `failed`), then complete the `TaskResult` (WebSocket notify).
3. **Lookup / API status**: `preview_status` is `"ready"` only when every raster frame has a complete image for the fingerprint; otherwise `"notready"`. `multiframe_previews` is only serialized when ready.

### How new styles update items
On style **create/update**, the client sends `raster_style_params` (write-only). The serializer saves the style, then calls `invalidate_and_enqueue_previews(instance)`:
- New fingerprint → new/regenerating preview rows for that style’s params.
- Existing fingerprints are left alone (no supersede); uniqueness is by fingerprint.
- Response returns `preview_status: "notready"` and omits `multiframe_previews` until generation finishes.
Also triggered after:
    - **Dataset conversion** — empty-params (`{}`) previews for each multiframe layer.
    - **Flood simulation** — sets style `raster_style_params` and regenerates sync so outputs are ready before the flood task completes.

## REST Endpoint Updates
- Prefetch helper `layer_queryset_with_previews()` loads raster frames + their preview rows in one go.
- **`LayerViewSet` / dataset `layers`**: use that queryset; `LayerSerializer` exposes `preview_status` and `multiframe_previews` (default fingerprint: default style params, else `{}`).
- **`LayerStyleViewSet`**: uses `LayerStyleWithPreviewsSerializer` with the same fields keyed to the style’s fingerprint; create/update accept `raster_style_params` and enqueue regeneration.
- Null preview fields are omitted from responses.

### Notification suppression
`suppress_task_notifications()` (`ContextVar`) silences `TaskResult` WebSocket pushes during sync contexts (ingest, sync conversion, sync preview generation) where no client is listening. Failed pushes are also best-effort (logged, never abort `save()`).

### `asynchronous=True` → `run_mode` ([Ruff / boolean-trap](https://docs.astral.sh/ruff/rules/boolean-default-value-positional-argument/))
Boolean default args for async vs sync were replaced with an explicit enum.:
- `TaskRunMode.ASYNC` / `TaskRunMode.SYNC` (`"async"` / `"sync"`)
- Call sites updated: `Dataset.spawn_conversion_task` / `convert_dataset`, `create_layers_and_frames`, preview enqueue helpers, flood/geoai analytics, ingest sample scripts.
Async preview jobs enqueue with `transaction.on_commit(...delay)`. Sync jobs run under `suppress_task_notifications()` via `.apply()`, matching the surrounding conversion/ingest mode (no Celery worker required for ingest).

## Client Updates — How the Client Gets Information
1. **Initial load**: layer/style API payloads include `preview_status` and (when ready) `multiframe_previews` (`url`, `width`, `height`, `bounds`). Selecting a layer copies those onto the active style and prefetches image URLs.
2. **Style save**: client posts `raster_style_params`, immediately marks local state `notready`, clears stale previews, and dismisses the map overlay so real tiles show while regenerating.
3. **Completion via WebSocket**:
   - Project-scoped regenerations → analytics WS (`analysis` store).
   - Conversion-time default previews (often no project) → conversion WS (`conversion` store).
   - Both call `framePreviewStore.onPreviewTaskComplete`, which re-fetches the layer or style and reattaches previews to selected copies still using that style.
4. **Map UX** (`framePreview` store + `framePreviewLayer` utils): show the current-frame PNG as a MapLibre image overlay, prefetch adjacent frames for scrubbing, then fade to real tiles once the tile source loads. Layers/legend panels indicate when a preview overlay is active. Style editing dismisses previews until edit mode ends.


uvdat/core/frame_previews/lookup.py
```diff
@@ -0,0 +1,102 @@
...
+    for frame in frames:
+        prefetched = getattr(frame, "_prefetched_objects_cache", None)
+        if prefetched is not None and "previews" in prefetched:
+            match = next(
+                (p for p in prefetched["previews"] if p.style_fingerprint == fingerprint),
+                None,
+            )
+            if match is not None:
+                result[frame.id] = match
+            continue
```
annehaley, line 39: Should the `continue` statement be nested inside `if match is not None:`? If the match is None, we should add to `missing_frame_ids`, right?

uvdat/core/frame_previews/preview_regeneration.py
```diff
@@ -0,0 +1,296 @@
...
+        .order_by("id")
+        .first()
+    )
+
+
+def previews_current_for_fingerprint(layer: Layer, fingerprint: str) -> bool:
+    """Return whether every frame already has a complete preview for this fingerprint."""
+    frames = layer.raster_frames()
+    if len(frames) <= 1:
+        return False
```
annehaley, line 54: There are a few places where you still have this manual check for the number of frames, but I think you can just use the `is_multiframe_raster` function you added.

uvdat/core/frame_previews/preview_regeneration.py
```diff
@@ -0,0 +1,296 @@
...
+
+    by_frame = previews_by_frame_id(frames, fingerprint)
+    return all(
+        (preview := by_frame.get(frame.id))
+        and preview.status == PreviewStatus.COMPLETE
+        and preview.image
+        and preview.style_fingerprint == fingerprint
+        for frame in frames
+    )
```
annehaley, line 63: This looks very similar to the logic in `frame_previews.lookup.preview_status_for_fingerprint`. I can see that this has an added condition, but I can't tell whether that difference is intentional. If not, could these two places share a helper function?

uvdat/core/frame_previews/preview_regeneration.py
```diff
@@ -0,0 +1,296 @@
...
+    project,
+    layer_style: LayerStyle | None,
+):
+    if project is not None:
+        return project
+    if layer_style is not None:
+        return layer_style.project
+    # Prefer a project that already includes this dataset so the analytics
+    # WebSocket (project-scoped) receives completion. Conversion-time tasks
+    # before a project link still fall back to the conversion channel.
```
annehaley, line 128: If a frame preview generation task occurs as a consequence of a dataset conversion, we should run the generation synchronously within the conversion task so the conversion only appears complete once the previews are ready too (the same way you did for the flood simulation task).

uvdat/core/frame_previews/preview_regeneration.py
```diff
@@ -0,0 +1,296 @@
...
+            )
+
+        transaction.on_commit(_enqueue_preview_task)
+        return
+
+    with suppress_task_notifications():
+        generate_frame_previews.apply(
+            args=(layer_id, fingerprint, params, result_id),
+            kwargs=task_kwargs,
+        )
```
annehaley, line 192: It looks like `dispatch_frame_preview_task` is only called in one place, so wouldn't it be easier to manage the completion of the transaction before calling this function? 

Also, it looks like the final `apply` would never be called, is that code path vestigial?

uvdat/core/tasks/dataset.py
```diff
@@ -139,15 +149,28 @@ def create_layers_and_frames(dataset, layer_options=None):  # noqa: C901, PLR091
...
                 )
 
+    # Default empty-params previews (no Project/LayerStyle required).
+    # Match the surrounding conversion mode: sync ingest must generate
+    # previews inline (no Celery worker required).
+    for layer in Layer.objects.filter(dataset=dataset):
+        if layer.is_multiframe_raster():
+            if task_result is not None:
+                task_result.write_status(f"Generating frame previews for layer {layer.name}...")
+            invalidate_and_enqueue_layer_previews(layer, {}, run_mode=run_mode)
```
annehaley, line 159: I think that if we generate previews within an async dataset conversion task, the preview generation should block completion of the dataset conversion. So I think we always want the run mode to be synchronous for this.

uvdat/core/tasks/frame_preview.py
```diff
@@ -0,0 +1,419 @@
...
+    style = LayerStyle.objects.filter(layer=layer, project=project, name="Default").first()
+    if style is None:
+        style = LayerStyle.objects.create(name="Default", layer=layer, project=project)
+        style.save_style_configs(DEFAULT_MULTIFRAME_RASTER_STYLE_SPEC)
+        style.raster_style_params = {}
+        style.save(update_fields=["raster_style_params"])
+    if layer.default_style_id is None:
+        layer.default_style = style
+        layer.save(update_fields=["default_style"])
+    return style
```
annehaley, line 199: Why do we still need to create a default style object?

web/src/components/sidebars/LayersPanel.vue
```diff
@@ -191,6 +192,21 @@ function setLayerActive(layer: Layer, active: boolean) {
                 </template>
                 {{ element.name }}
                 <template #append>
+                  <v-icon
+                    v-tooltip="
+                      framePreviewStore.isDisplayingPreview(element)
+                        ? 'Showing a low-resolution preview while full-resolution tiles load.'
```
annehaley, line 198: Maybe this tooltip should say "while default resolution tiles load" instead of "full-resolution", since the user won't see full-res while zoomed out.

web/src/components/sidebars/LayersPanel.vue
```diff
@@ -191,6 +192,21 @@ function setLayerActive(layer: Layer, active: boolean) {
...
+                        : undefined
+                    "
+                    icon="mdi-image-filter-hdr"
+                    size="small"
```
annehaley, line 202: I wonder if there's a more intuitive icon we can use. What about `mdi-relative-scale` or `mdi-image-size-select-large`?

web/src/components/sidebars/LegendPanel.vue
```diff
@@ -117,6 +118,16 @@ function getColorPropsCoverage(layer: Layer) {
       <v-list v-if="filteredLegend?.length" density="compact">
         <v-list-item v-for="layer in filteredLegend" :key="layer.id">
           {{ layer.name }}
+          <v-icon
+            v-if="framePreviewStore.isDisplayingPreview(layer)"
+            v-tooltip="
+              'Showing a low-resolution preview while full-resolution tiles load.'
+            "
+            icon="mdi-image-filter-hdr"
```
annehaley, line 126: Previous comments apply here too

annehaley (changes requested):
As with #397, I'll probably do a few rounds of review on this. My first pass is more focused on server-side, I'll test it locally and take a closer look at client-side on my next pass.

Could you rebase this and ensure that your commit messages have semantic prefixes? Without them, a release won't be made upon merge.

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: swapped to using `is_multiframe_raster` where possible in https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/0bc66a96eae5225c4b8cd9054d0bf7dc4a06b805

uvdat/core/frame_previews/lookup.py
BryonLewis: So this function is meant to use the prefetched previews if possible when they exist on the LayerFrame.  If they don't exist it will use a DB lookup to find the missing ones.
The higher level conditional `if prefetched is not None and "previews" in prefetched:` means that the current `frame` in the `frames: list[LayerFrame]` has the previews prefetched.  If a match is not found, it means that it doesn't exist in `RasterFramePreview` at all so it will be missing.  It would make no sense to then query RasterFramePreview.objects for the missing layer_frame_id and style fingerprint.

If the `frame` doesn't have prefetched `previews` we can then add it to the missing_frame_ids because some consuming function requires the dictionary mapping of layer_frame_ids to previews.

This is doing some overly complex stuff for niche conerns, I'm going to simplify it by not utilizing the prefetching and just have an additional query where it filters by the fingerprint.  I thing the simplification is worth having an additional small query/filter.

Simplified and remove the preview prefect in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/1a82b8e2240349e8d38ba893f1f9bc2a86001cd3

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: Good find, I didn't notice this during the process.
One was used as a conditional for internal gating to see if previews were already generated and current.  The other was returning a response to be used by the client to know if they should be displaying the previews or not.
I'll take the one out of `preview_regeneration.py` and move it into `lookup.py` but use the ready/notready as the base function.  I still think it makes sense to have an internal boolean return to know if previews are ready but the core logic doesn't need to be duplicated.


updated in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/dc92b28360b8f6310ee3d8688fa3719c495a2cc9

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: I don't think this is the correct referenced lines for this comment.  But I made it so that any preview generation task started by the `core/tasts/dataset.py` and specifically `create_layers_and_frames` will default to using `run_mode=Task.SYNC` inside of the `invalidate_and_enqueue_layer_previews` so it will run inside of the current task instead of starting a new task.

I also fixed some bad merge history in the `_dispatch_frame_preview_task` function that was duplicating some functions.
Done in : https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/9b3295783183d9ea1a432f573d932da4bc19400e

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: yeah there was some bad merging or some other issue when modifying code that duplicated sections of code.  This should be fixed in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/9b3295783183d9ea1a432f573d932da4bc19400e

uvdat/core/tasks/dataset.py
BryonLewis: I think this was related to another comment, I've updated to make sure this is alway run in TaskRunMode.SYNC so it will be a part of the conversion process.

This is updated in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/9b3295783183d9ea1a432f573d932da4bc19400e

uvdat/core/tasks/frame_preview.py
BryonLewis: This was leftover from before and never actually called.  The function is removed in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/5549cf2465762ba6eb0b7be009d208843eda3bd7

web/src/components/sidebars/LayersPanel.vue
BryonLewis: changed in the LayersPanel.vue and the LegendPanel.vue: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/4022d63db76870b83e3731583137e94c7063f175

web/src/components/sidebars/LayersPanel.vue
BryonLewis: swapped from mdi-image-filter-hdr to mdi-image-size-select-large: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/c8b6f108a49cb9ca36033915a4368f2ca319c70f

web/src/components/sidebars/LegendPanel.vue
BryonLewis: swapped from mdi-image-filter-hdr to mdi-image-size-select-large: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/c8b6f108a49cb9ca36033915a4368f2ca319c70f

changed in the text in LayersPanel.vue and the LegendPanel.vue: https://github.com/OpenGeoscience/geodatalytics/commit/4022d63db76870b83e3731583137e94c7063f175

uvdat/core/models/chart.py
```diff
@@ -47,5 +49,18 @@ def spawn_conversion_task(
...
+        # TODO: new line
+        pass
+
+    def rename_lines(self, new_names):
+        # TODO: rename lines
+        pass
+
+    def clear(self):
+        # TODO: clear
+        pass
```
annehaley: I think these lines were missed during a rebase; we've removed these.

uvdat/core/tasks/dataset.py
annehaley: The surrounding function no longer needs the `run_mode` arg since we always run preview generation in sync mode now.

uvdat/core/frame_previews/preview_regeneration.py
annehaley: Your commit addresses the code duplication, but I still think this function could be simpler if the transaction was committed before calling this function. You could use a `with transaction.atomic()` block in `preview_generation.invalidate_and_enqueue_layer_previews` and only call `_dispatch_frame_preview_task` once that context exits.

uvdat/core/frame_previews/preview_regeneration.py
```diff
@@ -0,0 +1,256 @@
...
+
+    return _TaskRunMode(run_mode)
+
+
+def layer_needs_previews(layer: Layer) -> bool:
+    return layer.is_multiframe_raster()
+
+
+def style_needs_previews(layer_style: LayerStyle) -> bool:
+    return layer_needs_previews(layer_style.layer)
```
annehaley: These two functions feel a bit redundant. Let's just use `is_multiframe_raster` directly and not obfuscate the logic.

uvdat/core/frame_previews/preview_regeneration.py
annehaley: I had left this comment here because running the preview generation task synchronously within the dataset conversion task means that we don't need to pick an arbitrary project to associate with the task. The dataset conversion tasks are intentionally not associated with any project. If we spawn a preview generation task because a style object is updated, we can use the style's project. 

Is there some other case wherein we need to pick an arbitrary project for the task result object? If not, I think `_resolve_preview_task_project` can be removed entirely and replaced with `layer_style.project || project`, which is allowed to be None.

uvdat/core/tasks/frame_preview.py
```diff
@@ -0,0 +1,400 @@
...
+FRAME_PREVIEW_DEFAULT_RESOLUTION_FRACTION = 1 / 8
+
+# Applied when ingest/API still creates a Default LayerStyle for vector config.
+DEFAULT_MULTIFRAME_RASTER_STYLE_SPEC = {
+    "default_frame": 0,
+    "opacity": 1,
+    "colors": [{"name": "all", "visible": True, "use_feature_props": True}],
+    "sizes": [{"name": "all", "zoom_scaling": True, "single_size": 5}],
+    "filters": [],
+}
```
annehaley: Is this also left over from creating default styles?

web/src/store/style.ts
```diff
@@ -399,11 +400,39 @@ function getVectorVisibilityPaintProperty(
...
 
   const mapStore = useMapStore();
   const projectStore = useProjectStore();
   const layerStore = useLayerStore();
   const networkStore = useNetworkStore();
+  const framePreviewStore = useFramePreviewStore();
+
+  function layerStyleKey(layer: Layer) {
+    return `${layer.id}.${layer.copy_id}`;
+  }
```
annehaley: We've needed a function like this for a while, so this is a helpful addition. Can you update other places that do this same thing to use this function instead? We can use this in `CompareLayerStyle.vue`, `LegendPanel.vue`,  `LayerStyle.vue` (in 2 places), and `map.ts`. It can also be used instead of `layerKey` in `framePreview.ts`, and there's another place in this file that could use it. Just search for `.copy_id}` and you'll see all the instances I mean.

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: Yeah that's right.  I'll use `with transaction.atomic()` in the `invalidate_and_enqueue_layer_previews` and also move the call to invalidate_and_enqueue_layer_previews out of the LayerStyleSerializer (where there was an implicit commit during save) and into the LayerViewSet where it happens after the serializer saves.

This is done in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/bbb928100b382550cb2b59d51796f162be5c6def

uvdat/core/tasks/dataset.py
BryonLewis: I've remove the run_mode arg from the `create_layers_fames` and `convert_dataset` both didn't require a run_mode anymore the only place.  `spawn_conversion_task` utilizes `run_mode` to determine if a task result should be created.  `invalidate_and_enqueue_layer_previews` also uses `run_mode` so when layerStyles are created/modified it will run in async for creation of new frame previews.

uvdat/core/models/chart.py
BryonLewis: removed in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/67a6e9ed712558c8a76b1d0ad57b79580fd296ef

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: I think they were leftover form the previous setup where the layerStyles was more directly connected to the preview and I had just removed the additional logic and kept the function names.  You're right it makes no sense to keep them as simple wrappers.

updated in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/c879494a615965ca96e691818cfa2cf1bebdef1f

uvdat/core/tasks/frame_preview.py
BryonLewis: Yeah, it was leftover from the previous version and was being used by tests.  I've remove this and the references in the tests in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/d8424685567bdfa9a8b863a3e7c725efda9ea4d5

web/src/store/style.ts
BryonLewis: I believe I found all instances, I also swapped a `mapStore.uniqueLayerIdFromLayer(layer)` to `layerStyleKey(layer)` as well.  All instances of creating that id with the copy_id should flow through style.ts and `layerStyleKey(layer) now.

This is updated in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/5479c16d52ffeccee6cb27b07f776ae60fd82e33

annehaley: Yeah that's better, but I notice that those places now import that function directly. For files that already import and use the style store, we can use `styleStore.layerStyleKey` instead of importing it separately, because the style store includes that function.

BryonLewis: I had a different mental model of how style.ts was being used.  I thought there were utility functions that weren't dependendent on reactivity that were in style.ts but not `useStyleStore`.  (Even if I did move layerStyleKey into the return value of useStyleStore 🤦)  But now I see that the all/most of the functions are returned in  useStyleStore().  Discussion for another PR is if there should be a split between styleUtils and styleStore.

Do you want me to clean up the other instance: https://github.com/OpenGeoscience/geodatalytics/blob/063272107151cb29ac9b0fc25e84f7319b1c2128/web/src/components/sidebars/LegendPanel.vue#L7 
I.E remove the export `colormapMarkersSubsample` from style.ts and force LegendPanel.vue it to use the styleStore?

BryonLewis: Where possible in https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/f45437d547a4c17587ce4a5485913d66318d6b5e I moved to using the `styleStore.layerStyleKey`, I also updated that one `colormapMarkersSubsample` to use the style store as well.

NOTE:  There was one location in map.ts that couldn't use the styleStore because the import happens in `useMapStore` and is used in a utility function `sourceIdFromLayerFrame` so an export of layerStyleKey needed to be used unless we restructure how that works.

annehaley: Ok, leaving that one instance in `map.ts` alone is fine, I just wanted to be sure we access it from the style store wherever possible. And thanks for doing the same with `colormapMarkerSubsample`.

BryonLewis:
Merging #458 improves/changes the following:

- Preview first loading - previews are loaded before the raster image to make sure that even when raster tiles a long time to come up a preview is shown relatively quickly
- `updateLayerStyles` now applies the new style_spec to all loaded frames, previously it was only the current frame so you would get a cached preview of the other layer style from the raster view (not the preview).  This was seen before as a flash of an old style right after the preview loaded and the raster tiles first displayed
- `clearPreviewsForStyleChange` - clears map preview overlays and invalidate cached `multiframe_previews` until the API reports that they are ready again.  This should make it so frame previews are all invalidated when a style is changed and relies on the raster tile until the previews are loaded again
- Frame Scrubbing Debounce - large lists would attempt to load the raster tiles immediately.  This has updated to give a 1000ms delay while scrubbing until it attempts to load the raster tile.
- Preview Indicator - solid blue for looking at a frame preview, flashing gray for generating previews, corresponding tooltip text has been updated as well
- `backfill_frame_previews` Management Command - a command that will create frame previews for any multiframe rasters that exist in the system.  It will first generate the None style previews and then generate any custom styles.  It does have a local `raster_style_params` generator so it doesn't need the client to do this
    - Helper `clear_frame_previews` management command that will clear all frame previews and reset the `raster_style_params` on any styles to be None.  Useful for testing the process for the `backfill_frame_previews`

uvdat/core/frame_previews/preview_regeneration.py
BryonLewis: The `layer_style.project || project` should be updated in: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/72c5c753bced79f4a11c7abe697a32826c4b5e95

BryonLewis:
I've done a few  more UI/UX updates to fix issues I've notced:
- There were still some minor remaining reordering issues that are now prevented:  https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/a09c6a83a8ffa4798fe6199979ddcf1146a5f3b7
- Reduce visual flickering during frame transitions by reusing existing MapLibre sources/layers when the previewURL hasn't changed: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/2c792afff68ad4643e727f98bf0d8283ab682084
- Fixes an issue when a style is deleted or switched, resets the stale preview overlays and refetches layer data: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/b693148d2265dcc1688652ecfb7be8b1aed0fb5d
- Fixed bug where if frame previews completed their task while the Style Editor was open they didn't update the frame previews once closed: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/14272873b2c5b93bdc17ec8f79ceb8d7b9523c38
- Fix scrubbing frame preview performance by making sure only the target layer updates instead of updating all layers: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/5e219f069acfcf5a2f3ab1eceef6c713116826db
- Handle NODATA in the source image for the previews and making it transparent (specifically flood simulation requires some additional checking and processing to make sure NODATA is set to transparent in the preview PNGs: https://github.com/OpenGeoscience/geodatalytics/pull/427/commits/a94c82896f1c4065994fa3096fab5c1226c87e63

annehaley (commented):
The client behavior is working a lot better now. I only noticed one thing on my recent test. 

I uploaded a multiframe raster dataset and it successfully created previews with the default style. Then I created some new styles with colormaps, and those successfully generated previews as well. Then I created a new style with the "None" color mode, going back to RGB. The preview generation task finished quickly because the first set of previews matched the new style, so the first previews could be reused. But the preview state icon stayed in the "notready" state, greyed out and flashing. I could scrub and see the previews, and the icon would correctly show its blue state when the previews were visible. But when it switched to tiles, the icon would go back to the "notready" state. Refreshing the page resolved this, so it's just a matter of updating the client state more accurately.

annehaley (commented):
If I upload a multiframe raster via the UI, it seems like the preview generation task starts before the conversion task is complete, leading to errors like this one:

Image

In this case, it seems like the conversion task was still in progress for the first 55 frames, so they all got this error that large image could not read the tiff (I assume because the file field is empty at that point). Then once the conversion task had saved the file field, the following 41 frames succeeded. Then both tasks exit (the preview generation one finishes before the conversion one). When I look at the code, it seems like the preview generation task should run synchronously at the end of the conversion task, so I'm not sure how this happened, but it happened to me twice.

I also noticed that in the case of this task failure, the preview state icon in the UI stays in the notready state indefinitely (even after refresh).

annehaley:
Actually, disregard the comment above. I restarted my celery container and could no longer replicate the issue. I guess I was working with an out-of-date version of the task.

annehaley (approved):
In my latest tests, I couldn't find any more unexpected behavior; everything I tried worked as expected. Thanks for fixing the issues I found previously. And thanks for persevering with such a large and complex PR. Very excited to have this functionality!

BryonLewis:
Thanks for putting up with all the changes and back-and-forth with regards to this feature!

PR #433, opened 2026-07-27
Use Redis pub/sub for Django Channels, to prevent errors when websockets disconnect

brianhelba:
See https://pypi.org/project/channels-redis/

### DIFF
--- a/uvdat/settings/base.py
+++ b/uvdat/settings/base.py
@@ -145,12 +145,14 @@ CORS_ALLOWED_ORIGIN_REGEXES: list[str] = env.list(
 _redis_url = env.url("DJANGO_REDIS_URL").geturl()
 CHANNEL_LAYERS: dict[str, dict[str, Any]] = {
     "default": {
-        "BACKEND": "channels_redis.core.RedisChannelLayer",
+        "BACKEND": "channels_redis.pubsub.RedisPubSubChannelLayer",
         "CONFIG": {
             "hosts": [
                 {
                     "address": _redis_url,
-                    "db": 1,
+                    # No "db" is set: the pub/sub layer only issues PUBLISH/SUBSCRIBE,
+                    # which are global to the Redis server and ignore the selected
+                    # logical database.
                 },
             ]
         },


annehaley:
Could you provide a bit more explanation on why we're making this change? I notice that the docs you linked explain that `RedisPubSubChannelLayer` is currently beta; can we be confident enough that this won't break with future changes?

brianhelba:
@annehaley This fixes https://kitware-data.sentry.io/issues/7615903301/?project=4510620385804288

In the old `RedisChannelLayer`, most of the time is spent awaiting [the `_brpop_with_clean` method](https://github.com/django/channels_redis/blob/48f0e0f1efeeea9ab985de5c3acbe5a67c03c2f4/channels_redis/core.py#L397-L399), which (in essence) awaits a blocking [the `BZPOPMIN` operation](https://redis.io/docs/latest/commands/bzpopmin/) and doesn't resolve until a message is available (or a Redis-level timeout occurs, and it cleanly loops again). However, if the network ever disconnects, a network `TimeoutError` is raised, which bubbles unhandled up to disrupt the WebSocket connection.

In the newer `RedisPubSubChannelLayer`, the code [calls `PubSub.get_message`](https://github.com/django/channels_redis/blob/48f0e0f1efeeea9ab985de5c3acbe5a67c03c2f4/channels_redis/pubsub.py#L314-L316) in a tight polling loop. Any connection failures will be [caught and the loop will retry](https://github.com/django/channels_redis/blob/48f0e0f1efeeea9ab985de5c3acbe5a67c03c2f4/channels_redis/pubsub.py#L326).

Ultimately, this is a bit speculative, so I want to test it empirically in a production-like environment to see if it works. While `RedisPubSubChannelLayer` is still marked in the docs as "beta", it was [added 5 years ago](https://github.com/django/channels_redis/pull/251) and has been maintained since. If we encounter problems, we can always revert. If it works, maybe this'll be a further positive signal to advocate that `channels_redis` mark `RedisPubSubChannelLayer` as stable.

annehaley (approved):
Alright, thanks for elaborating. That makes sense. Then we can merge this and test the production behavior.

PR #434, opened 2026-07-28
Fix raster tooltip

annehaley:
The raster value tooltip is one of the older visualization features, and it's been broken for a while. The `handleLayerClick` function was not registered properly on raster bounds layers. This PR fixes that bug so that the map tooltip will show a raster's pixel value when a raster layer is clicked.

More importantly, though, @brianhelba and I discovered that the API call to get the data for the raster tooltip was causing OOM errors in production when called on the Boston Orthoimagery layer, which is relatively large. The `get_image_data` function on the `RasterData` model accepted a resolution argument, and we were sending in a constant resolution value of `0.1`. But the resolution value was applied after getting the full-res data from large_image's `getRegion`. Loading the full orthoimagery layer into memory was exceeding Heroku's memory limits. 

This PR updates the `get_image_data` function to pass a `maxWidth` and `maxHeight` to `getRegion`. We define the maximum size as a constant on the server (rather than allowing the user to specify the size via the API and possibly exceed the memory limit, which would leave us vulnerable to DOS).

Lastly, this PR changes the format of the raster tooltip. Instead of assuming that we only want the value from the first band, we now display all band values as an array.

### DIFF
--- a/uvdat/core/models/data.py
+++ b/uvdat/core/models/data.py
@@ -15,6 +15,8 @@ from .dataset import Dataset
 from .file_item import FileItem
 from .querysets import ProjectQuerySet
 
+MAX_DATA_SIZE = 500
+
 
 class RasterData(models.Model):
     name = models.CharField(max_length=255, default="Raster Data")
@@ -29,17 +31,20 @@ class RasterData(models.Model):
     def __str__(self):
         return f"{self.name} ({self.id})"
 
-    def get_image_data(self, resolution: float = 1.0):
+    def get_image_data(self):
         with tempfile.TemporaryDirectory() as tmp:
             raster_path = Path(tmp, "raster")
             with raster_path.open("wb") as raster_file:
                 raster_file.write(self.cloud_optimized_geotiff.read())
             source = large_image.open(raster_path)
-            data, _data_format = source.getRegion(format="numpy")
-            data = data[:, :, 0]
-            if resolution != 1.0:
-                step = int(1 / resolution)
-                data = data[::step][::step]
+            data, _ = source.getRegion(
+                format="numpy",
+                output={
+                    "maxWidth": MAX_DATA_SIZE,
+                    "maxHeight": MAX_DATA_SIZE,
+                },
+            )
+            data = data[:, :, 0 : source.metadata.get("bandCount", 1)]
             return data.tolist()
 
 
--- a/uvdat/core/models/data.py
+++ b/uvdat/core/models/data.py
@@ -1,15 +1,11 @@
 from __future__ import annotations
 
 import json
-from pathlib import Path
-import tempfile
 
 from django.contrib.gis.db import models as geomodels
 from django.core.files.base import ContentFile
 from django.db import models
 from django.dispatch import receiver
-import large_image
-import numpy as np
 from s3_file_field import S3FileField
 
 from .dataset import Dataset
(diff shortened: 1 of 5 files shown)


manthey:
Are we really just getting the pixel values under the cursor?  Internally, large_image has a get-pixel method (which is essentially getting a 1x1 pixel region, though I think it returns the styled and unstyled pixel values).

annehaley:
We make one initial request to get a low-res copy of the data, and we keep that cached on the client side so that we don't have to make a get-pixel request for every mouse click on the layer.

BryonLewis:
Was it initially a hover tool to display the data?
The penalty of doing it for each intentional click for a single pixel is that more than the accuracy penalty of using a lower resolution version of the image (especially if the image has higher zoom levels)?

I went back and look at the PR history:
#11 was the inital hover state before maplibre that used the pixel under mouse
At some point it was changed to maplibre and I believe the hover implemention was kept
Then here I think it was swapped from a hover to a click: https://github.com/OpenGeoscience/geodatalytics/commit/f8b6d6e26fef67dd8fa17d99216c3ccc930463fb#diff-df09a306dbe4f98281804c6bc174d7113c8645e063ab49746e334149006bf8b9R8-R45

It may make sense to just utilize the utility that David is talking about in a request instead of having a cached representation of values to the pixels (that could be at a lower resolution)

annehaley:
Ok, I tried this out in [8c071d0](https://github.com/OpenGeoscience/geodatalytics/pull/434/commits/8c071d0a8f74342f7477be1cb56e3d90a3d51f87). Requests to get a single pixel value took about 5 seconds for the orthoimagery layer and 600 ms for the elevation layer. This is a noticeable wait, but we display "Fetching pixel value..." in the tooltip, so this wait time may be acceptable. I couldn't say whether this wait time is a worthwhile tradeoff for better pixel accuracy. If not, I can just undo this commit and we can go back to low-res cached data.

uvdat/core/models/data.py
```diff
@@ -29,18 +30,20 @@ class RasterData(models.Model):
     def __str__(self):
         return f"{self.name} ({self.id})"
 
-    def get_image_data(self, resolution: float = 1.0):
+    def get_pixel(self, lat, lng):
         with tempfile.TemporaryDirectory() as tmp:
             raster_path = Path(tmp, "raster")
             with raster_path.open("wb") as raster_file:
                 raster_file.write(self.cloud_optimized_geotiff.read())
             source = large_image.open(raster_path)
```
BryonLewis: If this is a COG is there some way we can use GDAL/large-image to read it using range requests instead of downloading and writing the file locally for each pixel check?

BryonLewis (changes requested):
question about using range requests for pixel checking vs downloading the file and opening in large_image.

uvdat/core/models/data.py
annehaley: I looked into how `django-large-image` does it a bit more, and the base class we use for the RasterData ViewSet actually has a canonical way to get the tile source ([here](https://github.com/girder/django-large-image/blob/8f6c6aa8247e6161eaff99f93927cfd9259acbc4/django_large_image/rest/base.py#L99)). I believe loading the tile source like this will allow us to make range requests like you suggest. After the change in [8be059b](https://github.com/OpenGeoscience/geodatalytics/pull/434/commits/8be059b118413b0f0079db2a9de9efcc97146a1e), the response time for a pixel request on the orthoimagery layer was reduced to <100ms (compared to 5s).

BryonLewis: This fix works by using the already downloaded version of the COG (The tile serving downloads it locally) instead of downloading it each time so you should see a significant impact on the access time for the pixel request.  We may want to discuss a separate PR that moves this fully into the range request access instead.  That should reduce the load times for the initial visibility of the Tiles as well as memory requirement when loading large raster data.

annehaley: Ok that makes sense. If we do figure out a way to do proper range requests, we may even want to upstream that to `django-large-image`; that seems like helpful base functionality.

BryonLewis (approved):
Good with this for now.  

I suggest another PR to look into utilizing `LargeImageVSIFileDetailMixin` in Django Large Image as well as an update to [Large-Image](https://github.com/girder/large_image/blob/master/large_image/tilesource/geo.py#L32) and [Django-Large-Image](https://github.com/girder/django-large-image/blob/main/django_large_image/utilities.py#L74) to update the `make_vsi` function for minIO support.  Mostly to update to using vsis3 instead of vsicurl for MinIO local development

PR #435, opened 2026-07-30
Remove redundant layer visibility controls in legend panel

annehaley:
I asked @faiza-a about the duplicate controls we currently have for layer visibility. We use checkboxes in the selected layers panel and eye icons in the legend panel; these controls are synced. Having two different controls for the same thing is redundant and possibly confusing. 
image

Faiza pointed out that we don't need the controls in the legend panel, so this PR removes them. Going forward, the legend panel will be a read-only summary of the state configured in the selected layers panel.


BryonLewis (approved):
👍

PR #436, opened 2026-07-31
Fix follower permissions

annehaley:
This PR fixes the project permission logic for followers. On the server-side, this means adding several checks to API endpoints that allow creation of objects associated with a project (analytics runs, layer styles, colormaps, and view states). On the client side, this means consolidating the logic for project permissions to a new computed variable `permissions` in the project store. Several components needed updates to conditionally render controls according to those permissions.

### DIFF
--- a/uvdat/core/rest/analytics.py
+++ b/uvdat/core/rest/analytics.py
@@ -92,14 +92,10 @@ class AnalyticsViewSet(ReadOnlyModelViewSet):
     )
     def run(self, request, project_id: int, task_type: str, **kwargs):
         project = Project.objects.get(id=project_id)
-        if (
-            not request.user.is_superuser
-            and project.owner() != request.user
-            and request.user not in project.collaborators()
-        ):
+        if not project.user_can_edit(request.user):
             return Response(
                 "You do not have permission to run analytics tasks in this project.",
-                status=400,
+                status=403,
             )
         analysis_type_class = next(
             iter(at for at in analysis_types if at().db_value == task_type), None
--- a/uvdat/core/rest/analytics.py
+++ b/uvdat/core/rest/analytics.py
@@ -96,6 +96,15 @@ class AnalyticsViewSet(ReadOnlyModelViewSet):
     )
     def run(self, request, project_id: int, task_type: str, **kwargs):
         project = Project.objects.get(id=project_id)
+        if (
+            not request.user.is_superuser
+            and project.owner() != request.user
+            and request.user not in project.collaborators()
+        ):
+            return Response(
+                "You do not have permission to run analytics tasks in this project.",
+                status=400,
+            )
         analysis_type_class = next(
             iter(at for at in analysis_types if at().db_value == task_type), None
         )
--- a/uvdat/core/rest/analytics.py
+++ b/uvdat/core/rest/analytics.py
@@ -25,10 +25,6 @@ class AnalyticsViewSet(ReadOnlyModelViewSet):
         url_path=r"project/(?P<project_id>[\d*]+)/types",
     )
     def list_types(self, request, project_id: int, **kwargs):
-        # TODO: remove this when analytics are ready to be shown to all users
-        if not request.user.is_superuser:
-            return Response([], status=200)
-
         serialized = []
         for analysis_type in analysis_types:
             if not analysis_type.is_enabled():
--- a/uvdat/core/rest/colormap.py
+++ b/uvdat/core/rest/colormap.py
@@ -15,14 +15,10 @@ class ColormapViewSet(ModelViewSet):
 
     def create(self, request, **kwargs):
         project = Project.objects.get(id=request.data.get("project"))
-        if not (
-            request.user.is_superuser
-            or project.owner == request.user
-            or request.user in project.collaborators()
-        ):
+        if not project.user_can_edit(request.user):
             return Response(
                 "You do not have permission to create colormaps in this project.",
-                status=400,
+                status=403,
             )
         serializer = ColormapSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
--- a/uvdat/core/rest/colormap.py
+++ b/uvdat/core/rest/colormap.py
@@ -1,11 +1,34 @@
 from __future__ import annotations
 
+from django.db import transaction
+import jsonschema
+from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet
 
-from uvdat.core.models import Colormap
+from uvdat.core.models import Colormap, Project
 from uvdat.core.rest.serializers import ColormapSerializer
 
 
 class ColormapViewSet(ModelViewSet):
     queryset = Colormap.objects.all()
     serializer_class = ColormapSerializer
+
+    def create(self, request, **kwargs):
+        project = Project.objects.get(id=request.data.get("project"))
+        if not (
+            request.user.is_superuser
+            or project.owner == request.user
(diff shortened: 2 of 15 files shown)


uvdat/core/rest/analytics.py
```diff
@@ -96,6 +92,15 @@ def list_results(self, request, project_id: int, task_type: str, **kwargs):
...
     def run(self, request, project_id: int, task_type: str, **kwargs):
         project = Project.objects.get(id=project_id)
+        if (
+            not request.user.is_superuser
+            and project.owner() != request.user
+            and request.user not in project.collaborators()
+        ):
+            return Response(
+                "You do not have permission to run analytics tasks in this project.",
+                status=400,
```
BryonLewis: would a [403 status](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/403) be more fitting for this?  It's technically not a bad request, they just don't have permissions.

This will most likely apply to other locations in the code where permissions  Response code appears.

uvdat/core/rest/colormap.py
```diff
@@ -1,11 +1,34 @@
...
 
 class ColormapViewSet(ModelViewSet):
     queryset = Colormap.objects.all()
     serializer_class = ColormapSerializer
+
+    def create(self, request, **kwargs):
+        project = Project.objects.get(id=request.data.get("project"))
+        if not (
+            request.user.is_superuser
+            or project.owner == request.user
```
BryonLewis: I don't see an `@property` decorator on the `Project` model so I assume this needs to be `project.owner()` instead of `project.owner`?  Might want to add test for this I think that may have caught the issue? (Update: I have another comment about using a utility function, if that is used I don't think you need to test each permission section individually)

uvdat/core/rest/layer.py
```diff
@@ -46,6 +46,16 @@ def get_queryset(self):
         return qs
 
     def create(self, request, **kwargs):
+        project = Project.objects.get(id=request.data.get("project"))
+        if not (
+            request.user.is_superuser
+            or project.owner == request.user
+            or request.user in project.collaborators()
+        ):
```
BryonLewis: This logic block is done multiple times in this PR, should it be refactored into a helper function.
Fixing the `project.owner` to `project.owner()` and then something like
```python
def user_can_edit_project(user, project: Project) -> bool:
    return (
        user.is_superuser
        or project.owner() == user
        or user in project.collaborators()
    )
```
Then in:
analytics.py
colormap.py
layer.py
view_state.py
you could just call `if not user_can_edit_project(request.user, project):`

web/src/components/projects/ProjectConfig.vue
```diff
@@ -430,7 +420,7 @@ watch(
                     <v-icon icon="mdi-content-save" />
                   </v-btn>
                 </div>
-                <div v-if="['owner'].includes(permissions[project.id])">
+                <div v-if="deleteAllowed">
```
BryonLewis: These changes make it so that icons appear for all projects (even ones you are only a follower on) when you have one selected in your list that you are owner/collaborator on.

I.E if you have two projects in the list and one is your own project and the other you are a follower on if you select your own project the follower project will have the pencil/edit icons on.  The endpoints are guarded but the edit and delete show up and don't respond with an error message if they fail.

web/src/components/sidebars/AnalyticsPanel.vue
```diff
@@ -77,8 +77,15 @@ const selectedInputs = ref<Record<string, any>>({});
 const inputSelectionRules = [(v: any) => (v ? true : "Input required.")];
 const additionalAnimationLayers = ref();
 const inputForm = ref();
+const runAllowed = computed(() => {
+  if (!projectStore.currentProject) return false;
+  return ["owner", "collaborator"].includes(
+    projectStore.permissions[projectStore.currentProject.id],
+  );
+});
```
BryonLewis, line 85: Lower down this gates the v-tabs for the Analysis then it has another gate on the 'Run New' tab but not the "View Existing".  If the user is a follower should they be able to see other user's results that have collabortator or owner permission?

web/src/store/project.ts
```diff
@@ -46,6 +46,27 @@ export const useProjectStore = defineStore("project", () => {
...
+          p.collaborators.map((u) => u.id).includes(appStore.currentUser.id)
+        ) {
+          perm = "collaborator";
+        }
+        return [p.id, perm];
+      }),
+    );
+    return ret;
+  });
+
```
BryonLewis, line 74: The type structure for `availableProjects` has a 'followers' map to it.  This is defaulting everyone to being a follower if they aren't an owner or collaborator or superuser.  Should it explicitly check the followers for the projects.  I know that if they have the project in the list they are assumed to have at least follower status but you have this information from the server should it be checked and could possibly provide an error if a user is getting a list of projects that they don't have at least follower status on.

BryonLewis (changes requested):

uvdat/core/rest/analytics.py
annehaley: Good point, changed to 403s in [20c7f21](https://github.com/OpenGeoscience/geodatalytics/pull/436/commits/20c7f218ef3d4bf634c336727efa543c3729edb7)

uvdat/core/rest/layer.py
annehaley: I agree a helper function is appropriate here. In [20c7f21](https://github.com/OpenGeoscience/geodatalytics/pull/436/commits/20c7f218ef3d4bf634c336727efa543c3729edb7) I put one on the Project model, so we can call `project.user_can_edit(request.user)`.

uvdat/core/rest/colormap.py
annehaley: Helper function added in [20c7f21](https://github.com/OpenGeoscience/geodatalytics/pull/436/commits/20c7f218ef3d4bf634c336727efa543c3729edb7) uses `owner()`

web/src/components/projects/ProjectConfig.vue
annehaley: Oh yeah I totally missed this. I hadn't realized that this is in a for loop where `project` is not necessarily the selected project. Fixed in [6b8d2f5](https://github.com/OpenGeoscience/geodatalytics/pull/436/commits/6b8d2f5d0352f37f6969fd8464d25b8402bc9686).

web/src/store/project.ts
annehaley: Yeah we previously just assumed that if a user received information about a project from the server, then they must have at least follower status. But to protect against any unexpected server behavior, I added an explicit check for follower permissions in [6b8d2f5](https://github.com/OpenGeoscience/geodatalytics/pull/436/commits/6b8d2f5d0352f37f6969fd8464d25b8402bc9686).

web/src/components/sidebars/AnalyticsPanel.vue
annehaley: Yeah, if a user is a follower, they won't see tabs, they just see the content of the "View Existing" tab, which will show runs created by others on the project. For example:
image

BryonLewis: My mistake, I didn't have any analyticis to test and didn't feel like running them.  From  quick look at the code I thought it was hiding both tabs (which it is) and the content (but it's showing the runs).

BryonLewis (approved):
Thanks for addressing my comments, looks good.

PR #439, opened 2026-07-31
Remove old Jupyter widget

annehaley:
In today's weekly meeting, we decided that we no longer need to support the old Jupyter widget, so this PR removes that legacy code.


BryonLewis (approved):
LGTM 👍

PR #440, opened 2026-07-31
Clean up Charts API

annehaley:
The Charts Viewset has some old non-functional endpoints that we no longer use, and the charts tasks has some old functions that were also unused, so this PR removes those. This PR also adds a convert endpoint to the charts viewset to match the convert endpoint on the datasets viewset.

### DIFF
--- a/uvdat/core/models/chart.py
+++ b/uvdat/core/models/chart.py
@@ -32,18 +32,20 @@ class Chart(models.Model):
 
         convert_chart_signature = convert_chart.s(self.id, conversion_options)
         if asynchronous:
-            convert_chart_signature.delay()
+            # Prevent circular import
+            from uvdat.core.models.task_result import TaskResult  # noqa: PLC0415
+
+            result = TaskResult.objects.create(
+                name=f"Conversion of Chart {self.name}",
+                task_type="conversion",
+                inputs={
+                    "chart_id": self.id,
+                    "conversion_options": conversion_options,
+                },
+                status="Initializing task...",
+            )
+            convert_chart_signature.delay(result_id=result.id)
+            return result
         else:
             convert_chart_signature.apply()
-
-    def new_line(self):
-        # TODO: new line
-        pass
-
-    def rename_lines(self, new_names):
-        # TODO: rename lines
-        pass
-
-    def clear(self):
-        # TODO: clear
-        pass
+            return None
--- a/uvdat/core/rest/chart.py
+++ b/uvdat/core/rest/chart.py
@@ -1,12 +1,11 @@
 from __future__ import annotations
 
-from django.http import HttpResponse
 from rest_framework.decorators import action
 from rest_framework.response import Response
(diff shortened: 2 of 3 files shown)


uvdat/core/models/chart.py
```diff
@@ -32,18 +32,21 @@ def spawn_conversion_task(
 
         convert_chart_signature = convert_chart.s(self.id, conversion_options)
         if asynchronous:
-            convert_chart_signature.delay()
+            # Prevent circular import
+            from uvdat.core.models.task_result import TaskResult  # noqa: PLC0415
+
+            result = TaskResult.objects.create(
+                name=f"Conversion of Chart{self.name}",
```
BryonLewis:
```suggestion
                name=f"Conversion of Chart {self.name}",
```
probably want a space in there?

uvdat/core/models/chart.py
```diff
@@ -32,18 +32,21 @@ def spawn_conversion_task(
...
+                task_type="conversion",
+                inputs={
+                    "chart_id": self.id,
+                    "conversion_options": conversion_options,
+                },
+                status="Initializing task...",
+                creator=self.owner(),
```
BryonLewis: should this be `self.project.owner()`?  I don't think that chart has a direct owner, I didn't see anything in the model elsewhere.

uvdat/core/rest/chart.py
```diff
@@ -29,26 +28,8 @@ def files(self, request, **kwargs):
...
-        return HttpResponse(status=200)
-
     @action(detail=True, methods=["post"])
-    def clear(self, request, **kwargs):
+    def convert(self, request, **kwargs):
         chart = self.get_object()
-        self.validate_editable(chart, chart.clear)
-        return HttpResponse(status=200)
+        chart.spawn_conversion_task(conversion_options=request.data.get("conversion_options"))
+        return Response("Spawned conversion task", status=200)
```
BryonLewis: I don't think this is used in the client anywhere so a response change won't affect anything.  The dataset viewset also has a post action for conversion.  That returns a ResultTaskSerializer, should these be aligned?

BryonLewis (changes requested):

uvdat/core/models/chart.py
annehaley: Nice catch, I added that space and amended the existing commit.

uvdat/core/models/chart.py
annehaley: Oh yeah, this is a vestige of copying the `spawn_conversion_task` function on `Dataset`. I just removed that line, we set the result's `creator` in the api endpoint instead (with `request.user`).

uvdat/core/rest/chart.py
annehaley: Yeah, we should make those two `convert` endpoints similar. I changed this response to use the `TaskResultSerializer`.

BryonLewis (approved):
👍

PR #443, opened 2026-08-05
Remove non-functional icons

annehaley:
The Charts, Networks, and Analytics panels have some extra icons that don't do anything. They were intended to communicate the type of the objects in each list, but since they are shown among icons that have other functions (description tooltip and details dialog), it could be confusing that these icons don't do anything. Besides, the items in these lists all have the same type as indicated by the panel they're in.

This PR removes those unnecessary icons.


BryonLewis (approved):
👍

PR #444, opened 2026-08-06
Additional fixes for Charts

annehaley:
I was writing some example code for the user documentation that demonstrates how the charts API can be used, and I found a few more places that the charts API needed fixes.

1. If uploading a fileitem for a chart object, the fileitem creation endpoint should not require a dataset to be specified.
2. The conversion options passed from the `convert` endpoint are a JSON payload, and we need to parse that string.

Additionally, I realized that the x range controls in the charts panel would be far more intuitive as a range slider, rather than a start numeric input and a range numeric input.


uvdat/core/rest/chart.py
```diff
@@ -32,7 +34,7 @@ def files(self, request, **kwargs):
     def convert(self, request, **kwargs):
         chart = self.get_object()
         result = chart.spawn_conversion_task(
-            conversion_options=request.data.get("conversion_options")
+            conversion_options=json.loads(request.data.get("conversion_options"))
```
BryonLewis: Two things with this:
Is the conversion_options a required parameter for spawn_conversion_task?
json.loads(None) will through an error, should the error
Spawn conversion task in chart.py has `conversion_options` default to None, but then it calls `convert_chart` which seems to require some sort of conversion_options, specifically on getting labels, datasets and palettes.

uvdat/core/rest/file_item.py
```diff
@@ -25,30 +26,14 @@ def get_queryset(self):
...
-            metadata=request.data.get("metadata", {}),
-            index=request.data.get("index", 0),
-            dataset=dataset,
-            file=file_key,
-        )
-        serializer = self.get_serializer(file_item)
-        return Response(serializer.data, status=201)
+        except (signing.BadSignature, TypeError) as e:
+            raise ValidationError("Invalid file key") from e
+        serializer.save(file=file_key)
```
BryonLewis, line 36: I understand why this was done to allow uploading of the chart without requiring a dataset.  This does allow a user to upload orphaned fileItems.  I.E. FileItems that aren't connected to a chart or a dataset and therefore don't have any connection to a project I think.  So there isn't a way to delete them execept manually deleting the fileItem explicitly (which may be difficult based on permissions because I think permissions trickle down from the project level).

web/src/components/SliderNumericInput.vue
```diff
@@ -130,7 +130,7 @@ function updateRange(lower: number, upper: number) {
         :max="props.max"
         :step="props.step"
         :precision="precision"
-        :style="{ width: getInputWidth() }"
+        :style="{ width: getInputWidth(props.rangeModel[1]) }"
```
BryonLewis: would these make more sense to be the max of the rangeModel?
Right now the input boxes will change sizes has you go through the digits.  I.E each time you increase the number of digits the input box resizes causing the slider to adjust in size as well.  Attempt something with a range of 0-10000.

BryonLewis (changes requested):
I used some minor modifications of the scripts in the documentation.  Would it be useful to include some sort of those scripts in the ./scripts folder, maybe a uv/click option where you can do something like `uv run --script create_chart.py --project-id 12 --name "Example Chart"` then prompts for username/password (hidden password field) and it creates and binds a chart.

uvdat/core/rest/chart.py
annehaley: Yes, it is a required parameter. I added a check in [79e7ea4](https://github.com/OpenGeoscience/geodatalytics/pull/444/commits/79e7ea4f06ccaa3afa77f60e64ee4de5d84bafee) to make this requirement explicit. The only reason it's a kwarg in `spawn_conversion_task` is to avoid violating Ruff's FBT002 rule.

uvdat/core/rest/file_item.py
annehaley: I added a database constraint in [62e3bf8](https://github.com/OpenGeoscience/geodatalytics/pull/444/commits/62e3bf8d7f5a24747baef5c2153e51fb22cc60c5) which ensures that every `FileItem` must have a dataset or a chart relationship.

As a note to myself, I'll need to rebase this or #445, depending on which merges first. They both have migration files now, so I'll regenerate the migration file for whichever one we merge second.

web/src/components/SliderNumericInput.vue
annehaley: Ok, changed in [0e6a250](https://github.com/OpenGeoscience/geodatalytics/pull/444/commits/0e6a250c66491272201fad3cdde43e360d48ff17)

annehaley:
>  Would it be useful to include some sort of those scripts in the ./scripts folder, maybe a uv/click option where you can do something like `uv run --script create_chart.py --project-id 12 --name "Example Chart"` then prompts for username/password (hidden password field) and it creates and binds a chart.

I could add this script, but I don't think the use case is common enough that anyone would use it. It would probably be limited to the file contents approach for chart creation, because I think it would be cumbersome to pass manual chart data via the command line. I'll add the script if you think you'd use it, but otherwise I think the examples in the documentation will be sufficient.

BryonLewis:
> I could add this script, but I don't think the use case is common enough that anyone would use it. It would probably be limited to the file contents approach for chart creation, because I think it would be cumbersome to pass manual chart data via the command line. I'll add the script if you think you'd use it, but otherwise I think the examples in the documentation will be sufficient.

Completely up to you.  I just mentioned it because to test the charting functionality I either needed to ingest the boston flood data (which can be large), or find the documentation and create a project and modify/use it to load charts that have a numeric x-axis and regular axis.  This could be another PR:  another option would be to include in sample_dataset a lightweight test dataset that implements a lot of the functionality so it can easily be ingested and used for testing features/functionality.

annehaley:
> This could be another PR: another option would be to include in sample_dataset a lightweight test dataset that implements a lot of the functionality so it can easily be ingested and used for testing features/functionality.

The `test` use case in `sample_data` includes the tide levels chart, but I could swap that out for one of the hydrographs from the `boston_floods` use case if we want something smaller.

BryonLewis:
I'm sorry I completely missed that there is a tests folder in sample_data.  I usually focused on the higher level JSON files and I have some of my own untracked folders in there (for  sequential rasters and other sample data) and it got lost in the noise.  You can disregard my comment about the script then.

BryonLewis (approved):
👍

PR #445, opened 2026-08-07
Unauthenticated mode

annehaley:
This PR adds a new boolean field `allow_unauthenticated` to the `Project` model. Marking this field as true will allow users to see the project without logging in. This PR allows access to GET endpoints if not authenticated, adds new cases to our existing test suite to cover unauthenticated requests, and updates the client code such that an unauthenticated user may use the app but any controls that would result in database changes are hidden.

Lastly, I added some UI components to be shown when starting to use unauthenticated mode. We show an initial welcome dialog that lets the unauthenticated user know that the application is limited when not logged in. That dialog has two options: "Show me a tutorial" and "Explore without tutorial". Clicking the latter will just close the welcome dialog and let the user continue. Clicking the former will start a series of popups that explain the purpose of various interface sections. The video linked below demonstrates the tutorial behavior.

https://drive.google.com/file/d/15--wSmSsIObnmPlq1O6OO5xyZXhpEmtm/view?usp=sharing


web/src/components/projects/AccessControl.vue
```diff
@@ -104,7 +108,7 @@ onMounted(() => {
             class="mx-3 user-circle"
             :ripple="false"
           >
-            <span v-if="project.owner.first_name || project.owner.last_name">
+            <span v-if="project.owner.first_name && project.owner.last_name">
```
BryonLewis: I know signup requires first/last names, but as a model I don't think they are both required and as a typescript type they aren't.
Should the Letter Icon be enabled even if one is found?
Also should it auto capitalize (just in testing I have a user called 'test test' and I saw tt in the circle.  I assume this was meant to be more like Google/Notion/Monday and how they display a user badge being the initials?

Image

web/src/components/projects/AccessControl.vue
BryonLewis: Now that I'm looking at it this same item appears in the Owner, Collaborator and Followers section, should it be abstracted out into a separate component.  Maybe in another PR?
Also that would probably address these differences:

Image

Image

My bryon.lewis@kitware.com is my super user account and you notice the owner will default to a user icon, but followers and others won't because of differences in how they are displayed.

web/src/components/projects/ProjectConfig.vue
BryonLewis, line 400: Just a suggestion, I'm not sure if it fits in this PR but maybe some indicator that the Project  allows unauthenticated access?  The Project list view in Django admin doesn't have it, so you just need to either open an incognito window and look at the unauthenticated items or know which ones are enabled.

web/src/App.vue
```diff
@@ -12,13 +12,144 @@ import {
...
+  },
+];
+const currentTutorialStep = computed(() => {
+  if (
+    showTutorialStep.value == 0 ||
+    showTutorialStep.value > tutorialSteps.length
+  )
+    return undefined;
+  return tutorialSteps[showTutorialStep.value - 1];
+});
```
BryonLewis: Couple of things:
I think this could warrant it's own separate location/file for this logic, the config and component section.

This is a very specific thing that could probably we moved into a future PR or we can probably ignore it.
1. Create a dataset and share it with unauthenticated and then create a view state where the panels are all moved around
2. use incognito mode to go in and start the tutorial, during the tutorial you can load the view state and it will move all of the panels around.  Now the tutorial panels don't align with the information panels (I don't expect most users to do this)
3. If you Refresh the page (Even hard refresh) it is still at localhost:8080/view/2 even though it displays localhost:8080.  You'll see the splash screen and if you click on Continue without loggin in it still thinks you're using localhost:8080/view/2 so it will start with the panels all moved arround.  A refresh makes it seem like you went to a different URL in the bar but internally it is still loadling like it is at subpath `/view/2` so it loads the state.

uvdat/core/tests/test_analytics.py
```diff
@@ -13,10 +14,14 @@
 )
 
 
+@pytest.mark.parametrize(
+    "client",
+    [lf("api_client"), lf("authenticated_api_client")],
+)
 @pytest.mark.django_db
-def test_rest_list_analysis_types(user, authenticated_api_client, project):
+def test_rest_list_analysis_types(user, client, project):
```
BryonLewis: Through testing I think the Anyltics types and projects are open to anyone.  I though that anayltics are tied to a project and that project would have it's access control or "allow_unauthenticated".  Most things I think go through the project to determine access but these endpoints I think don't.

uvdat/core/access_control.py
```diff
@@ -25,21 +30,33 @@ def has_object_permission(self, request, view, obj):
...
-        # Get all projects user has access to
-        user_projects = get_objects_for_user(
-            klass=models.Project, user=request.user, perms=perms, any_perm=True
-        )
+        permitted_projects = Project.objects.all()
+        if request.user.is_anonymous:
+            permitted_projects = permitted_projects.filter(allow_unauthenticated=True)
+        elif not request.user.is_superuser:
+            permitted_projects = get_objects_for_user(
+                klass=permitted_projects,
```
BryonLewis, line 38: should the `permitted_projects` also include the allow_unauthenticated filter?  Currently a signed in user that is not super_user does not have access to allow_unauthenticated to view unless it is specificfly given to them?

uvdat/core/models/project.py
```diff
@@ -19,6 +19,7 @@ class Project(models.Model):
     default_map_center = geo_models.PointField()
     default_map_zoom = models.FloatField(default=10)
     datasets = models.ManyToManyField(Dataset, blank=True)
+    allow_unauthenticated = models.BooleanField(default=False)
```
BryonLewis, line 22: I know there was a discussion about only allowing super users to edit this.  I don't think there is any guard againts a non-user who has edit permissions from being able to edit this through a PATCH request.  We don't have any UI but it could possible be done.

uvdat/core/access_control.py
```diff
@@ -25,21 +30,33 @@ def has_object_permission(self, request, view, obj):
...
 
 class DatasetGuardianPermission(GuardianPermission):
     def has_object_permission(self, request, view, obj):
         if request.user.is_superuser:
             return True
 
+        if request.user.is_anonymous:
+            return (
+                request.method == "GET"
+                and obj.project_set.filter(allow_unauthenticated=True).count()
```
BryonLewis: should this use `exists()` like other locations in the access_control instead of counting all of them?

BryonLewis (changes requested):
Just some minor things found, some of them may make sense in a subsequent PR.

While we have some guarding on database mutations we may want to make an explicit decisions on what kind of data leakage is acceptable for an unauthenticated user.  Projects obviously are guarded but things like the analysis tasks/types for other items aren't.

web/src/components/projects/AccessControl.vue
annehaley: Good idea, I moved the user profile display to its own component so we can be sure that all three sections have a consistent appearance ([1014c38](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/1014c387393b70a9f5eb94fcddcc55b5bd59e6c9)). 

While revisiting this display, I remembered that @brianhelba had suggested that we should not show emails (they should be private), so I removed that subtitle row from the user profile component.

web/src/components/projects/AccessControl.vue
annehaley: The new `UserProfile` component in [1014c38](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/1014c387393b70a9f5eb94fcddcc55b5bd59e6c9) handles capitalization and consistent name(s) display. If a user doesn't have a first or last name, we display "Unnamed User" with a generic user icon. This only happens when we create a user without using the signup form, though.

web/src/components/projects/ProjectConfig.vue
annehaley: Ok, in [04fb05d](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/04fb05d268b10fc003e5b89e5979aafb47ba180f) I added an alert banner above the tabs on the project configuration page:
image

BryonLewis: If it's not too much trouble you think it could added to the Django Admin List Projects view as another column?

web/src/App.vue
annehaley: I moved the tutorial logic & template to its own component, `TutorialPopup`, in [a82d783](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/a82d78302f7b79c67dedf644287d58df9bbffae2).

Then in [cbc639f](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/cbc639ffbf1f93705d8bf74c8a952fc5cfcc067f) I moved the logic to a new store, `tutorialStore`, so that we can avoid loading a view state if we're showing the welcome dialog or tutorial popup. The view state is loaded after exiting the tutorial or clicking "Continue without tutorial".

uvdat/core/tests/test_analytics.py
annehaley: The `TaskResult` objects are tied to a project, but the Analysis Types aren't db objects, they just represent the presence of enabled tasks. Maybe for unauthenticated mode we could just return task types where at least one task run exists on a project that allows unauthenticated?

I implemented this idea in [4c0be78](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/4c0be781e6a1130c68f85c2bcc8712fbd7bb2e9f), lmk your thoughts.

web/src/components/projects/AccessControl.vue
brianhelba: Yes, this isn't just a suggestion, is an important privacy issue. You cannot be exposing users' email addresses to other untrusted users.

uvdat/core/access_control.py
annehaley: Yeah, I had assumed that logged-in users wouldn't want to see the demo, but I don't think that's a valid assumption anymore. Changed in [af99d79](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/af99d797705f86777e1b7b37f4243864e670d963)

uvdat/core/models/project.py
annehaley: I added a protection in the serializer validation in [7f9664b](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/7f9664b8788b62c72083b7b1414cb5d7f3b9e862)

uvdat/core/access_control.py
annehaley: Yep. changed in [9b68b7f](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/9b68b7fdee1903dced0186f00ad6d7d122143105)

web/src/components/projects/ProjectConfig.vue
annehaley: No trouble at all. Added in [dc12964](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/dc12964062010d69f71c2b412f934c8801d0c1e7). Sorry I missed that part of your original suggestion.

web/src/components/projects/AccessControl.vue
annehaley: I made sure to remove other instances of the email field in [268c8f2](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/268c8f2b93aa92b773964066f6758774eeb6141f). The user serializer no longer includes the field, so they can't be obtained from the API either.

brianhelba: What about `username`? I believe that contain the email address too.

We could reconfigure user registration to request a separate `username` from `email` for users, or we should treat `username` as equivalently private information.

brianhelba: For reference, this concern was originally discussed in: https://github.com/OpenGeoscience/geodatalytics/pull/354#discussion_r2949846062

annehaley: Good point, we'll make usernames private, too. Removed them in [6f3d13b](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/6f3d13b4f191286124ba23f680a1f3a60ad670c8).

BryonLewis: because you allow configuration of permissions with followers/collaborators we may want to do what Brian suggest with requesting a separate `username` from `email` for the users.
The latest updates I think would only allow a project owner to give access by first/last name which can be duplicated so a unique identifier like a username may be necessary?

brianhelba: Indeed, granting access by first/last name would even allow impersonation, which is a potential security risk.

This fundamentally depends on the type of sharing model:
* If the application will be deployed as a "walled garden", where registration is limited to trusted / approved users, then users can see email addresses (of their ostensible colleagues) to further manage sharing
* If the application is deployed with open registration to the public:
  * and only trusted (i.e. hand-picked) administrators will manage sharing, then such administrators can also see email addresses
  * and peer-to-peer sharing of resources is allowed, then you should enable distinct usernames (which is the precise reason that "usernames" conceptually exist, as they enable unique identification, while protecting the email address that backs the account)

annehaley: We do have public registration and anyone can manage access on a project they create, so usernames would make sense as the unique identifier. 

We currently use [`resonant_utils.allauth.FullNameSignupForm`](https://github.com/kitware-resonant/django-resonant-utils/blob/master/resonant_utils/allauth.py#L15), but I can make a new signup form class that adds username to that. I'm guessing we would need to manually update the existing users on the prod instance who already have their emails as their usernames, but that's mostly Kitware folks anyway. Should we also switch to logging in via username instead of email? We would need to undo some [stuff set by resonant](https://github.com/kitware-resonant/cookiecutter-resonant/blob/4061634d8803f62c1b8ca000c8c2beda8d9ab50e/django-resonant-settings/resonant_settings/allauth.py#L34). 

If the approach we want isn't already part of resonant, should we consider an upstream contribution? This seems like a pretty standard use case.

brianhelba: You shouldn't need to change the sign up form at all. That's just gathering 2 extra fields that the database model already supports.

The particular login fields are handled by Allauth. I'll follow up with more specific recommendations.

annehaley: Ok, I made a separate issue to track this: #446. We can continue this discussion there and address this in a separate PR. At least with hiding usernames and emails for now, we are safe enough to add the unauthenticated mode.

uvdat/core/rest/serializers.py
```diff
@@ -93,6 +94,12 @@ def to_internal_value(self, data):
             data["default_map_center"] = Point(center[0], center[1])
         return data
 
+    def validate_allow_unauthenticated(self, value):
+        request = self.context.get("request")
+        if value and not request.user.is_superuser:
+            raise ValidationError("Only superusers can allow unauthenticated access to a project.")
```
BryonLewis: what about a PermissionDenied instead of a ValidationError?
https://docs.djangoproject.com/en/6.0/ref/exceptions/#permissiondenied

annehaley: You're right, that is more appropriate here. Changed in [628f69d](https://github.com/OpenGeoscience/geodatalytics/pull/445/commits/628f69d3af121cc8a8c5b1580b7db9675da6dbba).

